PR #413, opened 2026-04-27
fix: action type reads "kind" key instead of wrong "type" key

Muizzkolapo:
## Summary
- Fix `execution_events.py`, `error_handling.py`, and `generator.py` reading the wrong config key (`"type"` instead of `"kind"`), causing action type tracking to always emit empty/default values
- Sibling pattern in `generator.py` (4 occurrences) also fixed — action kind comparisons for tool vs LLM classification were silently broken

## Verification
- `pytest` — 6027 passed, 2 skipped
- `ruff format --check` — clean
- `ruff check` — clean (1 pre-existing unused import in unrelated file)


Muizzkolapo:
Duplicate of #408 which is already merged. Same files, same fix.

PR #490, opened 2026-05-03
docs: add record lifecycle guide

Muizzkolapo:
## Summary
- Adds `docs/guides/record-lifecycle.md` — comprehensive guide covering the record state machine, boundary resets, guard skips, terminal states, `_state_history` audit trail, and field categories
- Adds entry to guides index page
- Adapted from internal article, rewritten to match docs site style guide (approachable authority, progressive disclosure, concrete examples, mermaid diagrams)

## Verification
- Markdown follows Docusaurus conventions (frontmatter, mermaid blocks, admonitions)
- No code changes — docs only


Muizzkolapo (commented):
Two inaccuracies in the history examples:

**1. `"from": null` in the complete record example is misleading**

The `transition()` function sets `"from": from_state_raw` which is `record.get("_state")`. After boundary reset, `_state` is `"active"`, so `from` will be `"active"` not `null`. The `null` case only happens on the very first transition of a brand-new record before any `_state` has been set. The step-by-step example earlier in the guide correctly shows `"from": "active"`, but the final "Complete Record" example uses `"from": null` for every action — these should be `"from": "active"`.

**2. History entries include `timestamp` and `detail` fields**

The history entry format in the code is:
```python
{"timestamp": ..., "action": ..., "from": ..., "to": ..., "reason": ..., "detail": ...}
```

The examples omit `timestamp` and `detail`. The table in "The Audit Trail" section correctly lists both, but none of the JSON examples include them. Worth adding `timestamp` to at least the complete record example so readers know what to expect in real data.

Both are cosmetic — the conceptual explanation is solid.

PR #524, opened 2026-05-12
feat: per-subfolder seed overrides for staging directories

Muizzkolapo:
## Summary
- Detect `_seed_overrides.yml`/`.yaml` in staging subfolders during initial pipeline processing
- Override entries merge with workflow-level seed_path (override wins, unspecified keys inherit)
- Override files excluded from staging file processing in both `should_skip_item()` and `collect_files_from_upstream()`
- Unknown override keys rejected with `ConfigValidationError` listing valid keys
- Non-string override values rejected with `ConfigValidationError`
- Override file size checked before parsing (64KB limit)
- Path security enforced via existing `resolve_seed_path()` on override-specified file references
- Overrides scoped to initial action only; downstream actions use workflow defaults
- Config mutation prevented — full nested dict reconstruction, no shallow copy trap
- INFO-level logging when override is detected and applied
- Parsed overrides cached per directory for performance

## Verification
- 48 unit tests covering detection, validation, merging, config isolation, file filtering, and integration
- 87-check simulation covering both production-code assumptions and implementation contracts
- Config mutation verified with `id()` identity checks on `agent_config`, `context_scope`, and `seed_path`
- Cross-file bleed test: subfolder_a with overrides followed by subfolder_b without
- BUG-15 test: defaults-only seed_path with shared reference doesn't corrupt other actions
- All 6617 existing tests pass (6665 total with new tests)
- `ruff format --check` and `ruff check` clean


Muizzkolapo:
Closing — the sidecar override approach needs a design rethink. The core mechanism works but the UX (override files in staging, subfolder-based routing) doesn't match how operators actually want to switch between seed files. Will revisit with a cleaner design.

PR #562, opened 2026-05-19
fix: rerun retries COMPLETED_WITH_FAILURES actions

Muizzkolapo:
## Summary
- Actions with partial failures (`COMPLETED_WITH_FAILURES`) were preserved on rerun, causing the entire action to be skipped — failed records were never reprocessed unless the user explicitly used the `retry` CLI command
- Added `COMPLETED_WITH_FAILURES` to the `retryable` set in `reset_retryable()` so plain reruns reset these actions to PENDING

## Verification
- Updated existing test `test_preserves_completed_with_failures` → `test_resets_completed_with_failures` to assert the new behavior
- 6846 passed, 0 failures
- `ruff check` + `ruff format --check` clean


Muizzkolapo:
Reopening against integration/batch-online-unification

PR #600, opened 2026-05-22
fix: clear prior conflicting dispositions before writing terminal

Muizzkolapo:
## Summary

Merges PR #594 from `integration/v0.2.0-quality` to main.

When writing a terminal disposition (SUCCESS, FILTERED, etc.), DELETE any prior rows for that `(action_name, record_id)` pair first. Prevents SUCCESS + FAILED coexisting in the disposition table — the UNIQUE constraint is `(action_name, record_id, disposition)` not `(action_name, record_id)`, so both could persist and cause phantom `COMPLETED_WITH_FAILURES` on re-run.

## Verification

- Write FAILED then SUCCESS for same record → only SUCCESS row remains
- `get_failed_items` returns empty after SUCCESS overwrites FAILED


Muizzkolapo:
No longer needed — int branch merged main forward

PR #679, opened 2026-06-09
feat(storage): delta storage for target_data — 20x reduction

Muizzkolapo:
## Summary
- Each action stores only its own content namespace instead of the full accumulated record (O(N) instead of O(N²))
- Reconstruction happens transparently inside `read_target()` — consumers see identical `list[dict]`
- Delta logic lives in the abstract `StorageBackend` base class — future backends (Postgres, Mongo) get it for free
- 3 special-case write paths tagged `_delta_mode: "full"`: carry-forward, version correlation, FILE expansion
- Format version check prevents silent corruption on downgrade

## Architecture
- `write_target()` → concrete in base class, extracts delta, delegates to `_write_target_raw()`
- `read_target()` → concrete in base class, reconstructs from upstream deltas via `_read_target_raw_batch()`
- `_delta_mode` field (delta/first/full) stripped before returning to consumers

## Storage reduction
- qana_quiz (45 actions, 254 records): 440MB → ~22MB (20x)
- Write latency for final action: ~1s → ~20ms (50x)

## Safety
- FM7: reconstruction uses only abstract methods
- FM8: missing upstream → `_reconstruction_incomplete` flag + warning
- FM9: untagged multi-namespace records → warning
- FM10: format version check → `ConfigurationError` on old code reading delta DB
- Self-reviewed: fixed async_run path, corrupt metadata guard, cross-file early exit

## Verification
- 7455 tests pass, 2 skipped
- `ruff check` and `ruff format --check` clean
- Manual round-trip verified: write 3-action pipeline as deltas, read back full records with all namespaces

### DIFF
--- a/agent_actions/llm/batch/services/processing.py
+++ b/agent_actions/llm/batch/services/processing.py
@@ -379,6 +379,10 @@ class BatchProcessingService:
                 len(carry_records),
                 action_name,
             )
+            # Tag carry-forward records as full — they already have complete
+            # content from reconstruction and must not be delta-stripped.
+            for record in carry_records:
+                record["_delta_mode"] = "full"
 
         try:
             carry_path.unlink()
--- a/agent_actions/llm/batch/services/processing.py
+++ b/agent_actions/llm/batch/services/processing.py
@@ -379,8 +379,6 @@ class BatchProcessingService:
                 len(carry_records),
                 action_name,
             )
-            # Tag carry-forward records as full — they already have complete
-            # content from reconstruction and must not be delta-stripped.
             for record in carry_records:
                 record["_delta_mode"] = "full"
 
--- a/agent_actions/output/writer.py
+++ b/agent_actions/output/writer.py
@@ -132,8 +132,16 @@ class FileWriter(ProcessorErrorHandlerMixin):
 
             self.storage_backend.write_target(self.action_name, relative_path, data)
 
+            # Strip _delta_mode before disk write — it's a storage-layer
+            # internal that must not leak to filesystem files.
+            clean_data = [
+                {k: v for k, v in record.items() if k != "_delta_mode"}
+                if isinstance(record, dict) and "_delta_mode" in record
+                else record
+                for record in data
+            ]
             ensure_directory_exists(file_path, is_file=True)
-            atomic_json_write(file_path, data)
+            atomic_json_write(file_path, clean_data)
 
             return file_path.stat().st_size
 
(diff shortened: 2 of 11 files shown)


Muizzkolapo:
### Code review

Found 3 issues:

1. No tests for delta extraction, reconstruction, format version check, or preview reconstruction path. ~300 lines of new logic with zero new tests. (CLAUDE.md Gate 1: "New behavior must have new tests." quality-gates.md Gate 6: "For every happy path test, immediately write the corresponding failure/edge case test.")

https://github.com/Muizzkolapo/agent-actions/blob/d8f57f98accd68e1f6709f6b294eb4f2ae2479f8/agent_actions/storage/backend.py#L112-L400

2. FM9 multi-namespace warning fires as a false positive on every normal record at every action beyond the first. After `read_target` reconstructs full records upstream, every record passed to `write_target` at action N has N namespaces -- this is the normal data bus accumulation, not an untagged carry-forward. The warning will spam logs on every write in every multi-action pipeline, and the suggested fix ("tag as full") defeats delta storage entirely.

https://github.com/Muizzkolapo/agent-actions/blob/d8f57f98accd68e1f6709f6b294eb4f2ae2479f8/agent_actions/storage/backend.py#L237-L248

3. Guard-skip tombstone records from `build_skipped()` are not tagged `_delta_mode="full"`. These records have `content = {**upstream, action_name: None}` (multi-namespace) but flow through `write_target` without a full tag. `_extract_delta` strips all upstream namespaces, keeping only `{action_name: None}`. Carry-forward, correlation, and expansion records are tagged -- but guard-skip tombstones from `unified.py` / `result_collector.py` are not. (bug due to `record_helpers.py:build_tombstone` and `envelope.py:build_skipped` producing multi-namespace content without tagging)

https://github.com/Muizzkolapo/agent-actions/blob/d8f57f98accd68e1f6709f6b294eb4f2ae2479f8/agent_actions/processing/strategies/file_tool.py#L99-L108

Generated with [Claude Code](https://claude.ai/code)

Muizzkolapo:
### Code review (post-fix pass)

Found 1 issue:

1. `async_run()` still builds dependency graph using `_get_reachable_actions()` which silently produces empty dependency lists for versioned actions. The sync `run()` path was fixed in commit 86389e0b to use execution levels instead, with an explicit comment explaining why `_get_reachable_actions` fails for versioned actions (base-name mismatch). The async path was not updated. Any workflow using `async_run()` with versioned actions gets wrong reconstruction — the same production bug that was fixed for sync. (CLAUDE.md staff-thinking: "After fixing a pattern, grep for the same pattern elsewhere. Fix all instances.")

https://github.com/Muizzkolapo/agent-actions/blob/86389e0b80f836b23c7232a8f0a79632e384cc59/agent_actions/workflow/coordinator.py#L427-L438

Generated with [Claude Code](https://claude.ai/code)

PR #715, opened 2026-06-20
fix(prompt): render save and scope parsing surface their failures

Muizzkolapo:
## Summary
- Finding #7 (HIGH): `_save_failed_render` no longer swallows `OSError` on disk write. The function propagates the error; the single caller catches it, logs a warning, and still raises the original `ConfigurationError` for the YAML parse failure.
- Finding #8 (MEDIUM): `extract_action_names_from_template` now logs a `WARNING` on Jinja2 syntax errors instead of silently returning `set()`. The warning includes line number, error message, and template snippet. Return type unchanged to preserve caller compatibility.
- Sibling fix: `reference_extractor._extract_jinja_references` gets the same warning-on-syntax-error treatment.

## Verification
- ruff check + ruff format: clean
- pytest: 7554 passed (16 new regression tests)
- judge: claude-consensus 5/5 YES
- smoke: qanalabs online pass (52 actions, 487s)

PR #718, opened 2026-06-20
fix(prompt): render save and scope parsing surface their failures

Muizzkolapo:
## Summary

Contract 554 (Findings #7/#8) plus the cascading fixes required to make the integration tip green:

- **Finding #7 (HIGH):** `_save_failed_render` no longer swallows `OSError`. The caller logs a warning and still raises the original `ConfigurationError`, so disk-full / permission failures are visible instead of silently collapsing the debug-save path to `""`.
- **Finding #8 (MEDIUM):** `extract_action_names_from_template` now raises `TemplateSyntaxError` (via cherry-pick of PR #716). Callers (`scope_inference`, `_check_template_scope_coverage`, `resolution_service`, `_build_graph`) catch it per-action and record the failure in the validation result rather than letting it propagate as an empty set.
- **Smoke blocker (HIGH, framework):** `_resolve_client` was returning the lazy `"module:Class"` import string instead of the resolved class — `dict.setdefault` is a no-op when the key already exists, so the lazy entry was never replaced. Downstream `client.invoke(...)` then crashed with `'str' object has no attribute 'invoke'` on first LLM call for any external provider. Replaced with plain assignment + regression test covering all 10 registry entries.
- **Sibling pickup:** Pi-consensus correctly flagged the silent `TemplateSyntaxError` swallow at `validation/static_analyzer/reference_extractor.py:149`. That sibling lived behind PR #716 (already merged to main, not yet on integration). Cherry-picked it here so the integration tip carries both fixes together.

## What the consensus review surfaced (and how it's addressed)

Six rounds of pi-consensus on this branch. Every blocker was addressed with code + tests:

| Blocker | Fix |
|---|---|
| `_resolve_client` returned a string instead of a class | `tests/unit/llm/test_client_resolution.py` (9 cases) |
| Sibling swallow in `reference_extractor.py:149` | Cherry-picked #716 |
| Stale docstring on `extract_action_names_from_template` | Docstring rewritten to document the `Raises:` contract |
| Public cycle APIs (`to_dict`, `get_data_flow_summary`) crashed on cycles | Wrapped to return degraded result + `cycle_error` key |
| `WorkflowSchemaService.get_execution_order` silent fallback to dict-key order | Removed, raises `ValueError`; `to_dict` degrades |
| `WorkflowSchemaService.to_dict` crashed on cyclic workflows | Same wrap-and-degrade pattern |
| `_build_errors` reset wiped lazy-build errors | Reset moved into `_build_graph` after the early-return |
| `_check_template_scope_coverage` appended to consumed `_build_errors` | Append to local `errors` (returned to caller) |
| Syntax error in `_add_agent_node` dropped the entire action from the graph | Stub node created with partial regex requirements; error recorded after add |
| `_extract_from_template` lost simple-brace `{action.field}` references on Jinja syntax error | Regex pass moved before Jinja pass; partial requirements attached to the exception |
| `extract_from_workflow` discarded partial requirements | Now uses `partial_requirements` from the exception |
| Misleading 'prompt retrieval failed' on Jinja syntax errors in `scope_inference` | Split the handlers: dedicated `TemplateSyntaxError` warning with line number |
| Duplicate cycle errors from redundant Step-1b check | Removed; `StaticTypeChecker.check_all` already records the cycle |

## Verification

- ruff format + ruff check clean
- pytest: 7501 passed, 2 skipped (full suite)
- 30+ new regression tests across `tests/unit/prompt/`, `tests/unit/llm/`, `tests/services/`, and `tests/validation/static_analyzer/`
- Smoke: framework `'str' object has no attribute 'invoke'` bug fixed (the qanalabs smoke now executes 5 actions instead of failing at the first one). Remaining smoke failure is downstream in the qanalabs UDF `flatten_canonical_questions` — the LLM returned empty arrays from `extract_raw_qa_*`, which the UDF treats as a hard failure. That is content/model behavior, not framework code.

## Files changed

- `agent_actions/prompt/context/scope_parsing.py` (Finding #8 docstring + raise)
- `agent_actions/prompt/context/scope_inference.py` (dedicated `TemplateSyntaxError` handling)
- `agent_actions/prompt/render_workflow.py` (Finding #7)
- `agent_actions/llm/realtime/services/invocation.py` (`_resolve_client` smoke fix)
- `agent_actions/workflow/schema_service.py` (cycle propagation + `to_dict` degrade)
- `agent_actions/validation/static_analyzer/reference_extractor.py` (#716 + partial-requirements)
- `agent_actions/validation/static_analyzer/workflow_static_analyzer.py` (#716 + stub-node + dedup)
- `agent_actions/validation/static_analyzer/field_flow_analyzer.py` (#716 + cycle degrade)
- `agent_actions/validation/preflight/resolution_service.py` (#716 caller fix)
- `tests/...` (regression coverage for every fix above)

### DIFF
--- a/agent_actions/llm/realtime/services/invocation.py
+++ b/agent_actions/llm/realtime/services/invocation.py
@@ -66,8 +66,11 @@ def _resolve_client(model_vendor: str) -> Any:
                     "install_command": f"uv pip install {package}",
                 },
             ) from err
-        CLIENT_REGISTRY.setdefault(model_vendor, cls)  # atomic under CPython GIL
-        return CLIENT_REGISTRY[model_vendor]  # return whatever won the race
+        # Replace lazy "module:Class" string with the resolved class so
+        # subsequent lookups skip the import. Plain assignment, not setdefault:
+        # the key already exists (with a string), so setdefault would no-op.
+        CLIENT_REGISTRY[model_vendor] = cls
+        return cls
     return entry
 
 
--- a/agent_actions/prompt/context/scope_inference.py
+++ b/agent_actions/prompt/context/scope_inference.py
@@ -249,8 +249,29 @@ def infer_dependencies(
 
     try:
         raw_prompt = PromptFormatter.get_raw_prompt(action_config)
-        if raw_prompt:
+    except Exception as e:
+        logger.debug(
+            "Prompt retrieval failed for template inference on '%s': %s",
+            action_name,
+            e,
+            exc_info=True,
+        )
+        raw_prompt = None
+
+    if raw_prompt:
+        from jinja2.exceptions import TemplateSyntaxError
+
+        try:
             template_actions = extract_action_names_from_template(raw_prompt)
+        except TemplateSyntaxError as e:
+            logger.warning(
+                "Template syntax error during scope inference for '%s' (line %s): %s — "
+                "context_scope inference skipped; the syntax error will surface at render time.",
+                action_name,
+                e.lineno,
+                e.message,
+            )
+        else:
             # Only add template-referenced actions that are valid workflow actions
             valid_template_actions = template_actions & set(workflow_actions)
             if valid_template_actions - referenced_actions:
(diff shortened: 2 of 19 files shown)


Muizzkolapo:
Closing as redundant. Contract 554-prompt-silent-fallbacks (Findings #7/#8) was already fixed by PR #715 (prompt) and PR #716 (validation sibling), both merged to main. The only net-new bug this branch carried was the `_resolve_client` smoke regression introduced by PR #632 on integration — that's been spun out as a small focused PR: #719.

PR #719, opened 2026-06-20
fix(llm): _resolve_client returns class, not lazy import string

Muizzkolapo:
## Summary

PR #632 (LLM module cleanup) changed `_resolve_client`'s cache update from
plain assignment to `dict.setdefault` for "thread safety":

```python
CLIENT_REGISTRY.setdefault(model_vendor, cls)  # atomic under CPython GIL
return CLIENT_REGISTRY[model_vendor]            # return whatever won the race
```

`setdefault` is a no-op when the key already exists. The key IS pre-seeded with
the lazy `"module:Class"` import string, so `setdefault` never replaced it. The
following `return CLIENT_REGISTRY[model_vendor]` then handed back the lazy
string, and the first LLM call for any external provider crashed with:

```
'str' object has no attribute 'invoke'
```

This is the smoke-blocking bug every clone has been documenting as "pre-existing
LLM client issue" — it's been blocking the qanalabs online smoke at
`summarize_page_content` since PR #632 landed.

Fix: revert to direct assignment. The race the original comment hinted at is
benign — concurrent threads import the same module and bind to the same class,
so last-write-wins with identical values.

**Main is not affected.** PR #632 only landed on `integration/v0.2.0-quality`;
main still has the original direct assignment.

## Verification

- `pytest tests/unit/llm/test_client_resolution.py`: 9/9 pass (all 7 lazy
  vendors + 3 eager internal clients)
- `ruff check` + `ruff format --check`: clean
- After this fix the qanalabs online smoke gets past `summarize_page_content`
  (verified locally: 5 actions complete instead of failing at action 1)

### DIFF
--- a/agent_actions/llm/realtime/services/invocation.py
+++ b/agent_actions/llm/realtime/services/invocation.py
@@ -66,8 +66,13 @@ def _resolve_client(model_vendor: str) -> Any:
                     "install_command": f"uv pip install {package}",
                 },
             ) from err
-        CLIENT_REGISTRY.setdefault(model_vendor, cls)  # atomic under CPython GIL
-        return CLIENT_REGISTRY[model_vendor]  # return whatever won the race
+        # Replace lazy "module:Class" string with the resolved class so
+        # subsequent lookups skip the import. Plain assignment, not setdefault:
+        # the key already exists (with a string), so setdefault would no-op
+        # and the next return would hand back the lazy string — causing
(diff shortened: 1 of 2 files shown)


Muizzkolapo:
Closing — main already has the correct `CLIENT_REGISTRY[model_vendor] = cls` assignment. The bug only exists on integration/v0.2.0-quality (from PR #632), and integration will be reconciled with main some other way. No upstream fix needed.

PR #720, opened 2026-06-21
fix(cli): clearer UDF discovery/import errors + inspect help text

Muizzkolapo:
## Summary

Splits `UDFLoadError` handling into a dedicated formatter so users see the failing module, file, and an install hint instead of a leaked `<discovery>` sentinel or a generic configuration error.

## What changed

**`UDFLoadError` (agent_actions/errors/configuration.py)**
- New `DISCOVERY_SENTINEL: Final[str]` class attribute signals failures that aren't Python imports (missing user-code directory, path-derivation errors), so consumers render directory-appropriate messages instead of "fix your imports" advice.
- `__init__` now produces a directory-shaped message (`"UDF discovery failed: …"`) when the sentinel is used, so `str(exc)` and logs don't leak the internal token.

**`UDFLoadErrorFormatter` (new — agent_actions/logging/errors/formatters/udf_load.py)**
- Two render paths: directory failure (`title="UDF discovery failed"`) vs import failure (`title="Failed to load UDF module '…'"`).
- Extracts the missing module from `ModuleNotFoundError.name` typed attribute, walked via `__cause__` only — never `__context__`, which can surface unrelated errors from the call stack.
- Install hint warns that PyPI names differ from import names (yaml→PyYAML, cv2→opencv-python, bs4→beautifulsoup4) and tells the user how to handle the local-module case.
- Forwards `search_path` / `requested_path` enrichment from the pipeline into the rendered context.

**Chain hardening (formatters/translator.py + sibling formatters)**
- `UDFLoadErrorFormatter` registered ahead of function/configuration formatters; both siblings now defer via explicit `isinstance(exc, UDFLoadError)` guards so chain reordering can't silently regress UDF rendering.
- `ConfigurationErrorFormatter` no longer claims `UDFLoadError` implicitly via the `ConfigurationError` base class.
- `FunctionNotFoundFormatter` migrated from string-name matching to typed `isinstance` checks; substring fallback kept for untyped errors.

**Discovery hardening (input/loaders/udf.py)**
- `discover_udfs()` no longer crashes with `UnboundLocalError` when `Path.relative_to` fails on symlink/Windows quirks — surfaces a typed `UDFLoadError` routed through the discovery sentinel.

**Pipeline consistency**
- `config_pipeline._discover_udfs_from_path`, `validate_udfs.validate`, and `list_udfs.execute` all enrich `UDFLoadError` with `pipeline_stage` / `search_path` / `requested_path` so the formatter renders the configured location across every entry point, not just the per-file path.
- `validate_udfs._handle_load_error` delegates to `format_user_error` (markup disabled so brackets in messages aren't consumed as Rich markup) instead of hand-rolling Rich output.

**CLI help (cli/inspect_action.py)**
- `agac inspect action` and `agac inspect context` declare `metavar="ACTION_NAME"` and explicitly call out the positional argument (users were mistaking it for a `--action` flag). Examples preserved on separate lines via Click's `\b`.

## Test plan

- [x] `pytest tests/` (7621 passed)
- [x] `ruff format --check` + `ruff check`
- [x] `harness ci` + `harness review` (6 consecutive consensus passes)
- [x] `harness smoke` end-to-end against qanalabs-quiz-maker (last run hit Ollama 429 session limit; prior runs passed)
- [x] 100+ new regression tests across:
  - `tests/unit/error_formatters/test_udf_load.py` — formatter behavior, sentinel routing, chain integration, pipeline enrichment, `__cause__` chain walking
  - `tests/unit/error_formatters/test_function.py` — `UDFLoadError` defer
  - `tests/unit/error_formatters/test_configuration.py` — `UDFLoadError` defer
  - `tests/cli/test_list_udfs.py`, `tests/cli/test_inspect_field_flow.py` — CLI help + enrichment
  - `tests/unit/workflow/test_config_pipeline_stage.py`, `tests/validation/test_validate_udfs.py` — typed try/except enrichment
  - `tests/core/test_udf_loader.py` — `relative_to` `ValueError` wrapping

### DIFF
--- a/agent_actions/logging/errors/formatters/udf_load.py
+++ b/agent_actions/logging/errors/formatters/udf_load.py
@@ -1,0 +1,150 @@
+"""UDF discovery/import error formatter."""
+
+import re
+from typing import Any
+
+from agent_actions.errors import UDFLoadError
+
+from ..user_error import UserError
+from .base import ErrorFormatter
+
+_MISSING_MODULE_RE = re.compile(r"No module named ['\"]([^'\"]+)['\"]")
+
+
+def _first_non_none(*values: Any) -> Any:
+    """Return the first non-None value; unlike ``or``, empty strings / 0 / False are kept."""
+    for v in values:
+        if v is not None:
+            return v
+    return None
+
+
+class UDFLoadErrorFormatter(ErrorFormatter):
+    """Render UDF import failures with the module, file, and how to fix it.
+
+    UDF discovery imports every Python file under the user-code directory; a
+    single broken file blocks every command that loads the workflow. The
+    generic configuration formatter loses the module + file context that
+    ``UDFLoadError`` already carries, so this formatter claims those errors
+    before the configuration chain sees them.
+    """
+
+    def can_handle(self, exc: Exception, root: Exception, message: str) -> bool:
+        return isinstance(exc, UDFLoadError) or isinstance(root, UDFLoadError)
+
+    def format(
+        self, exc: Exception, root: Exception, message: str, context: dict[str, Any]
+    ) -> UserError:
+        # Read identity fields from the UDFLoadError directly: an outer wrapper
+        # that re-enriches these keys would otherwise clobber them via
+        # merge_exception_context (outer wins).
+        udf_exc = exc if isinstance(exc, UDFLoadError) else root
+        udf_context = getattr(udf_exc, "context", None) or {}
+
+        module = _first_non_none(udf_context.get("module"), context.get("module"), "<unknown>")
+        file = _first_non_none(udf_context.get("file"), context.get("file"))
+        error = _first_non_none(udf_context.get("error"), context.get("error"), message)
+        search_path = context.get("search_path")
+        requested_path = context.get("requested_path")
+
+        if module == UDFLoadError.DISCOVERY_SENTINEL:
+            return self._format_discovery_failure(file, error, search_path, requested_path)
+
+        return self._format_import_failure(module, file, error, exc, search_path, requested_path)
+
+    def _format_discovery_failure(
+        self,
+        path: str | None,
+        error: str,
+        search_path: str | None,
+        requested_path: str | None,
+    ) -> UserError:
+        ctx: dict[str, Any] = {}
+        if path:
+            ctx["file_path"] = path
+        if search_path:
+            ctx["search_path"] = search_path
+        if requested_path:
+            ctx["requested_path"] = requested_path
+        return UserError(
+            category="Configuration Error",
+            title="UDF discovery failed",
+            details=error,
+            fix=(
+                "Pass a valid user-code directory via `-u <path>`, or set "
+                "`tool_path` in your workflow config to a directory that exists."
+            ),
+            context=ctx or None,
+            docs_url="https://docs.runagac.com/user-defined-functions",
+        )
+
+    def _format_import_failure(
+        self,
+        module: str,
+        file: str | None,
+        error: str,
+        exc: Exception,
+        search_path: str | None,
+        requested_path: str | None,
+    ) -> UserError:
+        details = f"Python could not import the UDF module: {error}"
+        missing = self._extract_missing_module(error, exc)
+
+        where = "the UDF file shown above" if file else "the UDF file that triggered this"
+        if missing:
+            fix = (
+                f"Python could not find the module '{missing}'.\n"
+                f"  If it is a third-party dependency, install it "
+                f"(e.g. `uv add <package>` or `pip install <package>`) — "
+                f"note the PyPI package name may differ from the import name "
+                f"(yaml→PyYAML, cv2→opencv-python, bs4→beautifulsoup4).\n"
+                f"  If '{missing}' is a local module, fix the import path in "
+                f"{where}."
+            )
+        else:
+            fix = (
+                f"Fix the import error in {where}, then re-run.\n"
+                "  The exact failure is included in the Problem line."
+            )
+
+        fix += (
+            "\n\nUDF discovery imports every Python file under the user-code "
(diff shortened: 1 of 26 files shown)


agent_actions/logging/errors/translator.py
```diff
@@ -28,8 +29,16 @@ class ErrorTranslator:
...
+        # Chain ordering matters for safety: UDFLoadErrorFormatter MUST come
+        # before FunctionNotFoundFormatter and ConfigurationErrorFormatter so
+        # that UDFLoadError (a ConfigurationError subclass with a 'function-not-
+        # found'-shaped message in some cases) is rendered with the dedicated
+        # module/file/install-hint UX. The sibling formatters also have explicit
+        # isinstance(exc, UDFLoadError) guards so ordering isn't the only line
+        # of defense — see configuration.py and function.py.
         self.formatters: list[ErrorFormatter] = [
             YAMLSyntaxErrorFormatter(),  # Check YAML first (most specific)
+            UDFLoadErrorFormatter(),  # UDF import failures
```
Muizzkolapo: **Ordering leaves a hole: `YAMLSyntaxErrorFormatter` still steals UDF errors.** `UDFLoadErrorFormatter` is inserted here, but it sits *after* `YAMLSyntaxErrorFormatter`, which has no `UDFLoadError` exclusion guard (unlike the guards this PR added to `configuration.py` and `function.py`). A UDF that calls `yaml.safe_load()` on malformed YAML at import raises `yaml.YAMLError`, wrapped as `UDFLoadError(...) from e`; `extract_root_cause` returns the `YAMLError` as `root`, so `yaml.py:16` (`isinstance(root, yaml.YAMLError)`) claims it first and renders a generic "YAML syntax error" — dropping the module/file/search_path this feature exists to show. Either move UDFLoadErrorFormatter before the YAML formatter or add the same guard to `yaml.py`.

agent_actions/logging/errors/formatters/udf_load.py
```diff
@@ -0,0 +1,177 @@
...
+    @staticmethod
+    def _extract_missing_module(error_text: str, exc: Exception, root: Exception) -> str | None:
+        # Prefer the typed ModuleNotFoundError.name attribute — robust against
+        # message localization and reformatting. Walk __cause__ only (explicit
+        # `raise ... from ...` chain); do NOT walk __context__, which Python
+        # populates with whatever exception happened to be handled when this
+        # one was raised and could surface an unrelated ModuleNotFoundError
+        # from elsewhere in the call stack.
+        seen: set[int] = set()
+        for start in (exc, root):
```
Muizzkolapo: **`__cause__`-only guard is defeated by its own input.** The walk correctly avoids `__context__` when starting from `exc`, but it also seeds from `root` — and `root` is produced by `extract_root_cause` (utils/safe_format.py), which *does* fall through to `__context__`. So if the UDF error was raised while an unrelated `ModuleNotFoundError` was being handled (it lands in `__context__`), that module name leaks into `root` and the user is told to `pip install <unrelated-package>`. Consider extracting the name only from the `exc`/`__cause__` chain, not from the already-context-walked `root`.

agent_actions/logging/errors/formatters/udf_load.py
```diff
@@ -0,0 +1,177 @@
...
+        # would be clobbered. For these identity fields, prefer the
+        # UDFLoadError's own .context first so the formatter always renders
+        # the UDF's actual identity regardless of wrapping.
+        udf_exc = exc if isinstance(exc, UDFLoadError) else root
+        udf_context = getattr(udf_exc, "context", None) or {}
+
+        # Explicit None checks (not `or`) — UDFLoadError contracts the keys to
+        # non-empty strings, but an empty value would silently fall through `or`
+        # to the wrong source and mask the bug.
+        module = _first_non_none(udf_context.get("module"), context.get("module")) or "<unknown>"
```
Muizzkolapo: **`_first_non_none` is defeated by the trailing `or` here.** The helper + its comment exist to preserve a falsy/empty value so a contract violation surfaces — but `... or "<unknown>"` (and `... or message` on line 61) immediately re-swallow the empty string, so an empty `module`/`error` still falls through to the default. The helper only adds value at the `file` site (no trailing `or`); at these two sites it's equivalent to a plain `or`, making the rationale comment misleading. Either drop the `or` or drop the helper here.

agent_actions/workflow/config_pipeline.py
```diff
@@ -122,10 +122,33 @@ def _discover_udfs_from_path(path: str, project_root: Path | None, console: Cons
...
-        console.print(f"[cyan]\U0001f50d Discovering Tools in {abs_path}...[/cyan]")
+    if not (abs_path.exists() and abs_path.is_dir()):
+        # Pre-existing behaviour: workflows with optional tool_paths can list
+        # entries that aren't present in every environment, so silently skip
+        # missing directories at this layer. A typo'd path therefore won't
+        # raise — log it at DEBUG so it shows up under `--verbose` without
+        # spamming normal runs. discover_udfs's UDFLoadError.DISCOVERY_SENTINEL
+        # path remains reachable via direct invocations (e.g. validate-udfs)
+        # that bypass this guard.
+        logger.debug("Skipping UDF discovery for missing path: %s", abs_path)
```
Muizzkolapo: **Inconsistent surfacing of a missing tool_path.** A typo'd/missing `tool_path` returns 0 with a `debug`-only log, while `agac validate-udfs` and `agac list-udfs` raise `UDFLoadError` for the same missing directory. So `agac run` silently loads zero tools and fails later with a confusing "function not found", whereas validate/list give a clear error. Pre-existing behavior, but this PR restructured this function and is explicitly about unifying UDF-error UX — worth aligning (at least `warning`, ideally surface it).

PR #724, opened 2026-06-26
feat(logging): suppress noisy third-party loggers at framework init (spec 555 step 1)

Muizzkolapo:
## Summary
First slice of spec 555 (logging UX, dbt-standard CLI output). SDK/HTTP libraries (httpx, urllib3, openai, anthropic, ollama, groq, cohere, google.genai, googleapiclient, sentence_transformers, transformers, pypdf) are clamped to WARNING on every `LoggerFactory.initialize()` so framework-internal HTTP and SDK chatter stops leaking onto the CLI.

Users can override the curated defaults — or set per-logger levels on any other logger — via `LoggingConfig.module_levels`. Validation runs at apply time: non-string keys, non-string values, invalid level names, and overrides on `agent_actions` or its children are all rejected with stderr warnings rather than crashing or silently dropping. The `agent_actions` root must stay at DEBUG so every event reaches `events.json` — `module_levels` guards this.

## What's in this PR
- `_THIRD_PARTY_NOISY_LOGGERS` + `_apply_logger_levels()` in `LoggerFactory.initialize`
- `module_levels` typed as `dict[str, Any]` to honestly reflect the YAML/programmatic surface; runtime validation handles correctness
- `VALID_LOG_LEVELS` lifted to module-level constant; `from_environment` routes through `_validate_log_level` so WARN/WARNING aliasing is consistent across YAML, env vars, and `module_levels`
- `level_map` adds `CRITICAL → EventLevel.ERROR` so a user setting `default_level: CRITICAL` gets errors-and-above, not a silent fallback to INFO
- `LoggerFactory.reset()` restores third-party loggers to NOTSET so tests don't leak global state
- `LoggerFactory.set_level` / `set_debug` deleted — zero callers, contradictory semantics (silently broke `events.json` invariant)
- New test file `tests/logging/events/test_third_party_suppression.py` with 25+ cases covering clamp, overrides, invalid inputs, agent_actions guard, reset isolation, force-reinit idempotency
- Docs (`docs/api/logging.md`) and module manifest updated

## Deferred (flagged by pi-consensus, out of scope for step 1)
- **YAML wiring for `module_levels`**: `LoggingConfig.from_project_config` parses `module_levels` correctly but the CLI doesn't yet route project YAML through it. Pre-existing gap (`from_project_config` has zero production callers — entire YAML-based config path is inert). Should land as its own PR.
- **Deprecation shims for removed `set_level`/`set_debug`**: pi-consensus flagged removal as a breaking API change. Both methods had zero internal/test callers and the broken semantics (silently truncating `events.json`) made them dangerous to leave in place. If external callers exist, this is intentionally surfaced as `AttributeError` at upgrade time.
- **`module_levels` env var (`AGENT_ACTIONS_LOG_MODULE_LEVELS`)**: not in spec 555 step 1 scope; programmatic API covers the path.

## Verification
- `harness ci` against `origin/main`: 7679 passed
- 6 rounds of pi-consensus review against the integration branch; remaining feedback is the three deferred items above (documented), plus a recurring type-tightening preference (`dict[str, Any]` → narrower) where loose typing is the honest match for the YAML/programmatic input surface

## Spec
`specs/backlog/555-logging-ux-dbt-standard.md` step 1 of 5. Subsequent slices (audit framework-internal warnings, build dbt-style CLI renderer, route errors through events, cascade collapsing) land separately.

### DIFF
--- a/agent_actions/logging/_MANIFEST.md
+++ b/agent_actions/logging/_MANIFEST.md
@@ -22,8 +22,8 @@ factories, filters, formatters, and the event-driven error/reporting plumbing.
 | `config.py` | Module | Dataclasses that capture project logging, handler, and formatting defaults. | `logging` |
 | `FileHandlerSettings` | Class | File handler configuration used by the factory. | `logging` |
 | `LoggingConfig` | Class | Central logging configuration builder with `from_project_config`/`from_environment`. | `logging` |
-| `factory.py` | Module | `LoggerFactory` that wires together configuration, filters, and handlers. Registers four handler types: Console, `events.json` (all levels), `errors.json` (ERROR-only, output_dir runs only), and `run_results.json`. On init, clamps noisy third-party SDK/HTTP loggers (httpx, urllib3, openai, anthropic, ollama, groq, cohere, google.genai, googleapiclient, etc.) to WARNING and applies user-supplied `logging.module_levels` overrides. | `logging` |
-| `LoggerFactory` | Class | Manage logger creation, level setting, and debug state. | `logging` |
+| `factory.py` | Module | `LoggerFactory` that wires together configuration, filters, and handlers. Registers four handler types: Console, `events.json` (all levels), `errors.json` (ERROR-only, output_dir runs only), and `run_results.json`. On init, clamps noisy third-party SDK/HTTP loggers (httpx, urllib3, openai, anthropic, ollama, groq, cohere, google_genai, googleapiclient, etc.) to WARNING and applies user-supplied `logging.module_levels` overrides; pre-init levels are snapshotted and restored by `reset()`. | `logging` |
+| `LoggerFactory` | Class | Manages logger creation, third-party suppression, and the `events.json` bridge. | `logging` |
 | `filters.py` | Module | Custom filters (e.g., `RedactingFilter`) to sanitize sensitive payloads. | `logging` |
 | `formatters.py` | Module | Formatter helpers such as `JSONFormatter` used across services. | `logging` |
 
--- a/agent_actions/logging/_MANIFEST.md
+++ b/agent_actions/logging/_MANIFEST.md
@@ -22,7 +22,7 @@ factories, filters, formatters, and the event-driven error/reporting plumbing.
 | `config.py` | Module | Dataclasses that capture project logging, handler, and formatting defaults. | `logging` |
 | `FileHandlerSettings` | Class | File handler configuration used by the factory. | `logging` |
 | `LoggingConfig` | Class | Central logging configuration builder with `from_project_config`/`from_environment`. | `logging` |
-| `factory.py` | Module | `LoggerFactory` that wires together configuration, filters, and handlers. Registers four handler types: Console, `events.json` (all levels), `errors.json` (ERROR-only, output_dir runs only), and `run_results.json`. | `logging` |
+| `factory.py` | Module | `LoggerFactory` that wires together configuration, filters, and handlers. Registers four handler types: Console, `events.json` (all levels), `errors.json` (ERROR-only, output_dir runs only), and `run_results.json`. On init, clamps noisy third-party SDK/HTTP loggers (httpx, urllib3, openai, anthropic, ollama, groq, cohere, google.genai, googleapiclient, etc.) to WARNING and applies user-supplied `logging.module_levels` overrides. | `logging` |
 | `LoggerFactory` | Class | Manage logger creation, level setting, and debug state. | `logging` |
 | `filters.py` | Module | Custom filters (e.g., `RedactingFilter`) to sanitize sensitive payloads. | `logging` |
 | `formatters.py` | Module | Formatter helpers such as `JSONFormatter` used across services. | `logging` |
@@ -31,7 +31,7 @@ factories, filters, formatters, and the event-driven error/reporting plumbing.
 
 | Symbol | File | Interaction | Config Key |
 |--------|------|-------------|------------|
-| `LoggingConfig.from_project_config()` | `agent_actions.yml` | Reads | `logging`, `logging.level`, `logging.file` |
+| `LoggingConfig.from_project_config()` | `agent_actions.yml` | Reads (not wired into CLI yet — programmatic only) | `logging`, `logging.level`, `logging.file`, `logging.module_levels` |
 | `LoggingConfig.from_environment()` | `.env` | Reads | `AGENT_ACTIONS_DEBUG`, `AGENT_ACTIONS_LOG_LEVEL`, `AGENT_ACTIONS_LOG_FORMAT`, `AGENT_ACTIONS_LOG_DIR` |
 | `LoggerFactory.initialize()` | `agent_io/target/events.json` | Writes | — |
 | `LoggerFactory.initialize()` | `agent_io/target/errors.json` | Writes | — |
--- a/agent_actions/logging/factory.py
+++ b/agent_actions/logging/factory.py
@@ -26,7 +26,7 @@ _THIRD_PARTY_NOISY_LOGGERS: tuple[str, ...] = (
     "ollama",
     "groq",
     "cohere",
-    "google.genai",
+    "google_genai",
     "googleapiclient",
     "sentence_transformers",
     "transformers",
@@ -42,6 +42,9 @@ class LoggerFactory:
     _root_logger_name: str = "agent_actions"
     _event_manager: EventManager | None = None
     _run_results_collector: RunResultsCollector | None = None
+    # logger name → its level before the framework first touched it; used by reset()
+    # to restore pre-init state and by force-reinit to drop loggers no longer managed.
+    _original_logger_levels: dict[str, int] = {}
 
     @classmethod
     def initialize(
@@ -196,9 +199,8 @@ class LoggerFactory:
 
     @classmethod
     def _apply_logger_levels(cls, config: LoggingConfig) -> None:
-        """Clamp third-party loggers to WARNING; apply user module_levels overrides."""
-        for name in _THIRD_PARTY_NOISY_LOGGERS:
-            logging.getLogger(name).setLevel(logging.WARNING)
+        """Clamp noisy loggers + apply module_levels; restore prior level for loggers no longer managed."""
+        new_managed: dict[str, str] = {name: "WARNING" for name in _THIRD_PARTY_NOISY_LOGGERS}
 
         for name, level in config.module_levels.items():
             if not isinstance(name, str):
@@ -207,7 +209,6 @@ class LoggerFactory:
                     f"keys must be str, got {type(name).__name__}\n"
                 )
                 continue
-
             if name == cls._root_logger_name or name.startswith(cls._root_logger_name + "."):
                 sys.stderr.write(
                     f"agent_actions: ignoring logging.module_levels[{name!r}]={level!r}: "
@@ -215,17 +216,26 @@ class LoggerFactory:
                     "break events.json; use logging.level for CLI-visible levels.\n"
                 )
                 continue
-
-            normalized = str(level).upper()
-            if normalized == "WARN":  # Python alias for WARNING.
-                normalized = "WARNING"
-            if normalized not in VALID_LOG_LEVELS:
+            normalized = LoggingConfig.normalize_log_level(level)
+            if normalized is None:
                 sys.stderr.write(
                     f"agent_actions: ignoring logging.module_levels[{name!r}]={level!r}: "
-                    f"not one of {sorted(VALID_LOG_LEVELS)}\n"
+                    f"value must be one of {sorted(VALID_LOG_LEVELS)} "
+                    f"(got {type(level).__name__})\n"
                 )
                 continue
-            logging.getLogger(name).setLevel(normalized)
+            new_managed[name] = normalized
+
+        for name in list(cls._original_logger_levels):
+            if name not in new_managed:
+                logging.getLogger(name).setLevel(cls._original_logger_levels.pop(name))
+
+        for name in new_managed:
+            if name not in cls._original_logger_levels:
+                cls._original_logger_levels[name] = logging.getLogger(name).level
+
+        for name, level_str in new_managed.items():
+            logging.getLogger(name).setLevel(level_str)
 
     @classmethod
     def _setup_logging_bridge(cls) -> None:
@@ -298,7 +308,7 @@ class LoggerFactory:
 
     @classmethod
     def reset(cls) -> None:
-        """Reset factory state and unset third-party logger levels (for testing)."""
+        """Reset factory state and restore loggers we touched to their pre-init levels."""
         cls._initialized = False
         cls._config = None
         cls._event_manager = None
@@ -309,8 +319,9 @@ class LoggerFactory:
         for f in root_logger.filters[:]:
             root_logger.removeFilter(f)
 
-        for name in _THIRD_PARTY_NOISY_LOGGERS:
-            logging.getLogger(name).setLevel(logging.NOTSET)
+        for name, level in cls._original_logger_levels.items():
+            logging.getLogger(name).setLevel(level)
+        cls._original_logger_levels = {}
 
         from agent_actions.logging.core.manager import EventManager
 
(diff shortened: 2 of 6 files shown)


agent_actions/logging/factory.py
```diff
@@ -3,18 +3,37 @@
...
+_THIRD_PARTY_NOISY_LOGGERS: tuple[str, ...] = (
+    "httpx",
+    "httpcore",
+    "urllib3",
+    "openai",
+    "anthropic",
+    "ollama",
+    "groq",
+    "cohere",
+    "google.genai",
```
Muizzkolapo: 🔴 **Correctness — Gemini suppression is a no-op (wrong logger name).** The `google-genai` SDK logs under `google_genai.*` (underscore) — `google_genai.models`, `google_genai.files`, `google_genai._api_client` (verified in site-packages). `google.genai` (dot) is a *different* logger root and only catches the tiny `google.genai._interactions` subtree, so `google_genai.models` stays at INFO after the clamp. The wrong name is also copied into `docs/api/logging.md` and `_MANIFEST.md`. The parametrized test asserting `getLogger('google.genai').level == WARNING` is tautological — it re-reads this same `setLevel` and never touches a `google_genai` logger. **Fix:** use `google_genai` (or include both).

agent_actions/logging/factory.py
```diff
@@ -276,6 +309,9 @@ def reset(cls) -> None:
         for f in root_logger.filters[:]:
             root_logger.removeFilter(f)
 
+        for name in _THIRD_PARTY_NOISY_LOGGERS:
+            logging.getLogger(name).setLevel(logging.NOTSET)
```
Muizzkolapo: 🟠 **Correctness — `reset()` destroys logger levels it never owned.** This forces all 13 loggers to `NOTSET` instead of restoring their prior level (init at L201 overwrites to WARNING without snapshotting). A host app that did `getLogger('httpx').setLevel(ERROR)` before using the framework loses that level permanently after an init/reset cycle. And since `tests/conftest.py` has an **autouse** fixture calling `LoggerFactory.reset()`, every test teardown now mutates these 13 loggers' global state suite-wide. **Fix:** snapshot prior levels at init and restore them here.

agent_actions/logging/factory.py
```diff
@@ -173,6 +194,39 @@ def _register_handlers(
         manager.register(run_results)
         cls._run_results_collector = run_results
 
+    @classmethod
+    def _apply_logger_levels(cls, config: LoggingConfig) -> None:
+        """Clamp third-party loggers to WARNING; apply user module_levels overrides."""
+        for name in _THIRD_PARTY_NOISY_LOGGERS:
```
Muizzkolapo: 🟡 **Correctness (low) — stale overrides survive `force=True` re-init.** On `initialize(force=True)` this re-clamps only the 13 curated names plus the *new* config's `module_levels`. A logger set by a *previous* config's `module_levels` but absent from the new config is never reset and keeps its stale level. Narrow (programmatic `module_levels` + force), but the lifecycle should fully reflect the new config.

agent_actions/logging/factory.py
```diff
@@ -173,6 +194,39 @@ def _register_handlers(
...
+
+            if name == cls._root_logger_name or name.startswith(cls._root_logger_name + "."):
+                sys.stderr.write(
+                    f"agent_actions: ignoring logging.module_levels[{name!r}]={level!r}: "
+                    f"overrides on {cls._root_logger_name!r} or its children would "
+                    "break events.json; use logging.level for CLI-visible levels.\n"
+                )
+                continue
+
+            normalized = str(level).upper()
```
Muizzkolapo: 🟡 **Correctness (low) — value validation has no type guard (asymmetric with keys).** Keys are rejected via `isinstance(name, str)`, but values go straight through `str(level).upper()`. A non-string value whose `__str__` yields a level name (custom object/enum) is silently accepted and applied via `setLevel`, with no warning — the opposite of suppression. Reusing `config._validate_log_level` (which already type-guards) would close this; see the duplication comment below.

tests/logging/events/test_third_party_suppression.py
```diff
@@ -0,0 +1,200 @@
...
+        assert "module_levels" in captured.err
+        assert "httpx" in captured.err
+        assert "NOT_A_LEVEL" in captured.err
+        # Clamp ran AND invalid override was rejected → final state is WARNING,
+        # not DEBUG (would mean clamp didn't run) and not NOT_A_LEVEL (would
+        # mean the override sneaked past validation).
+        assert logging.getLogger("httpx").level == logging.WARNING
+
+    @pytest.mark.parametrize("bad_value", [True, 20, ["INFO"], None])
+    def test_non_string_module_level_rejected_at_runtime(self, bad_value, capsys):
```
Muizzkolapo, line 132: 🟡 **Test quality — name overclaims; can't catch the bug above.** Named `..._rejected_at_runtime` with a docstring saying non-string values are rejected, but the assertions only check the final level is WARNING and `'httpx'` is in stderr. `[True, 20, ['INFO'], None]` are rejected only because `str()` of each happens to fail membership — not by any type check. A regression that accepts a non-string stringifying to a valid level (the L219 hole) would still pass. Per the repo's 'name tests for what the assertions verify' rule, assert on type-rejection or rename.

agent_actions/logging/_MANIFEST.md
```diff
@@ -22,7 +22,7 @@ factories, filters, formatters, and the event-driven error/reporting plumbing.
 | `config.py` | Module | Dataclasses that capture project logging, handler, and formatting defaults. | `logging` |
 | `FileHandlerSettings` | Class | File handler configuration used by the factory. | `logging` |
 | `LoggingConfig` | Class | Central logging configuration builder with `from_project_config`/`from_environment`. | `logging` |
-| `factory.py` | Module | `LoggerFactory` that wires together configuration, filters, and handlers. Registers four handler types: Console, `events.json` (all levels), `errors.json` (ERROR-only, output_dir runs only), and `run_results.json`. | `logging` |
+| `factory.py` | Module | `LoggerFactory` that wires together configuration, filters, and handlers. Registers four handler types: Console, `events.json` (all levels), `errors.json` (ERROR-only, output_dir runs only), and `run_results.json`. On init, clamps noisy third-party SDK/HTTP loggers (httpx, urllib3, openai, anthropic, ollama, groq, cohere, google.genai, googleapiclient, etc.) to WARNING and applies user-supplied `logging.module_levels` overrides. | `logging` |
 | `LoggerFactory` | Class | Manage logger creation, level setting, and debug state. | `logging` |
```
Muizzkolapo: 🟠 **Conventions — stale manifest row.** This row still reads 'Manage logger creation, **level setting, and debug state**', but the PR removed `set_level`/`set_debug` (the methods that provided level-setting/debug-state). The adjacent module row *was* updated, so the manifest was touched but this row missed. Per CLAUDE.md 'Manifest Maintenance', an agent navigating via the manifest will call the removed methods → `AttributeError`. (L25 also still lists the wrong `google.genai` name — see the first comment.)

agent_actions/logging/factory.py
```diff
@@ -173,6 +194,39 @@ def _register_handlers(
...
+            if normalized == "WARN":  # Python alias for WARNING.
```
Muizzkolapo: 🟢 **Cleanup (reuse) — duplicated, already-drifted level normalization.** The `WARN→WARNING` + `VALID_LOG_LEVELS` membership here re-implements `config._validate_log_level` (config.py:73-78). The copies have **already diverged**: the config helper type-guards via `isinstance`, this path coerces via `str()` (that's the L219 bug). Calling the shared helper would DRY this *and* fix L219.

Muizzkolapo (commented):
## Code review — PR #724 (spec 555 step 1): suppress noisy third-party loggers

Max-effort review: 10 finder angles → adversarial 1-vote verification → gap sweep. 12 findings, posted inline and ranked.

**Top priority**
1. 🔴 **`google.genai` is the wrong logger name** — the SDK logs under `google_genai` (underscore), so Gemini suppression silently no-ops; the new clamp test passes tautologically. The wrong name is also in the docs and the manifest.
2. 🟠 **`reset()` clobbers logger state it doesn't own** — forces 13 loggers to `NOTSET` instead of restoring prior levels (breaks library embedders and mutates global state on every test via the autouse conftest reset).

**Lower priority:** force-reinit keeps stale overrides; value-side validation missing an `isinstance` guard (with a test that can't catch it); stale `_MANIFEST.md` row after `set_level`/`set_debug` removal; level-normalization + level-name-set duplication; hardcoded provider list duplicating the provider registries (the root cause of #1 — logger names guessed by convention); no config path to extend the denylist; repeated hand-rolled stderr warnings.

**Verified clear (not flagged):** removed `set_level`/`set_debug` have no callers and aren't exported; `module_levels`-from-YAML inertness is documented; the `OPENAI_LOG` stale-clamp is correct opt-in deference; validating the public `LoggingConfig(module_levels=…)` constructor is justified.

_Automated review via `/code-review` (max effort)._

agent_actions/logging/factory.py
```diff
@@ -3,18 +3,37 @@
...
     from agent_actions.logging.core.handlers import ContextDebugHandler
     from agent_actions.logging.core.manager import EventManager
     from agent_actions.logging.events.handlers import RunResultsCollector
 
 
+# Clamped to WARNING on init so SDK/HTTP chatter never reaches the CLI (spec 555).
+_THIRD_PARTY_NOISY_LOGGERS: tuple[str, ...] = (
+    "httpx",
+    "httpcore",
+    "urllib3",
```
Muizzkolapo, line 23: 🟢 **Altitude — hardcoded provider list duplicates the provider registries.** These provider roots restate the provider set already in `_VENDOR_PACKAGES` (invocation.py) and `_BatchClientRegistration` (batch_client_factory.py). Those store pip package names (`google-genai`) and logger-root ≠ package name — the exact convention-guessing that produced the Gemini bug above. Three places to keep in sync; consider each provider declaring its logger root once.

agent_actions/logging/factory.py
```diff
@@ -120,6 +140,7 @@ def _register_handlers(
             "WARN": EventLevel.WARN,
             "WARNING": EventLevel.WARN,
```
Muizzkolapo, line 144: 🟢 **Cleanup (reuse) — level-name set duplicated 3×.** These `level_map` keys, `config.VALID_LOG_LEVELS`, and the `LogLevel` Literal are three independent copies — and not identical (`VALID_LOG_LEVELS` omits the `WARN` alias this map carries). No single source of truth: add a level to one and the others silently disagree.

agent_actions/logging/factory.py
```diff
@@ -3,18 +3,37 @@
...
+from agent_actions.logging.config import VALID_LOG_LEVELS, LoggingConfig
 
 if TYPE_CHECKING:
     from agent_actions.logging.core.handlers import ContextDebugHandler
     from agent_actions.logging.core.manager import EventManager
     from agent_actions.logging.events.handlers import RunResultsCollector
 
 
+# Clamped to WARNING on init so SDK/HTTP chatter never reaches the CLI (spec 555).
+_THIRD_PARTY_NOISY_LOGGERS: tuple[str, ...] = (
```
Muizzkolapo, line 20: 🟢 **Altitude — denylist can't be extended via config.** `module_levels` only overrides severity of *known* loggers; there's no config/YAML way to *add* a logger to the clamped-by-default set. A newly noisy dependency requires a code edit + release — the treadmill the escape hatch was meant to avoid.

agent_actions/logging/factory.py
```diff
@@ -173,6 +194,39 @@ def _register_handlers(
...
 
+    @classmethod
+    def _apply_logger_levels(cls, config: LoggingConfig) -> None:
+        """Clamp third-party loggers to WARNING; apply user module_levels overrides."""
+        for name in _THIRD_PARTY_NOISY_LOGGERS:
+            logging.getLogger(name).setLevel(logging.WARNING)
+
+        for name, level in config.module_levels.items():
+            if not isinstance(name, str):
+                sys.stderr.write(
```
Muizzkolapo, line 207: 🟢 **Cleanup (low) — hand-rolled stderr warning repeated 4×.** `sys.stderr.write('agent_actions: ignoring …')` appears here and at L212/L223 and config.py:57, diverging from the package's `logger.warning` convention. The bootstrap-timing choice is defensible; a small shared helper would stop the prefix/wording drifting across 2 files.

agent_actions/logging/factory.py
```diff
@@ -173,6 +194,39 @@ def _register_handlers(
...
+                    f"agent_actions: ignoring logging.module_levels[{name!r}]={level!r}: "
+                    f"keys must be str, got {type(name).__name__}\n"
+                )
+                continue
+
+            if name == cls._root_logger_name or name.startswith(cls._root_logger_name + "."):
```
Muizzkolapo, line 212: 🟢 **Altitude (low) — guard sits apart from the invariant it protects.** This `agent_actions`-root carve-out enforces 'never override the framework tree', but the invariant (the tree stays DEBUG for `events.json`) is owned by `_setup_logging_bridge`, with an implicit apply-before-bridge ordering (L103 before L104). Two distant encodings can drift; both referencing `_root_logger_name` limits the risk.

Muizzkolapo:
## Response to review

Pushed `6b99ea40`. Addressed all correctness findings; deferred the architecture items as scope (none of them are bugs).

### Fixed

| # | Finding | Fix |
|---|---|---|
| 🔴 1 | `google.genai` → no-op for Gemini | `google_genai` (verified against `site-packages/google/genai/*.py` — all loggers use the underscore root). Fixed in `factory.py`, `_MANIFEST.md`, and `docs/api/logging.md`. |
| 🟠 2 | `reset()` clobbers logger state | New `_original_logger_levels: dict[str, int]` snapshots each logger's level the first time the factory touches it. `reset()` restores from that snapshot (host-app `httpx=ERROR` survives an init/reset cycle). Two regression tests cover this. |
| 🟠 3 | Stale `_MANIFEST.md` LoggerFactory row | Updated description; `set_level`/`set_debug` references gone. |
| 🟡 4 | Value validation missing `isinstance` guard | Routed through new `LoggingConfig.normalize_log_level(value: object) -> LogLevel \| None` which type-guards. The `str()` coercion is gone. |
| 🟡 5 | Test name overclaims; couldn't catch #4 | Added `test_value_that_stringifies_to_valid_level_still_rejected` with a class whose `__str__` returns `"INFO"` — asserts the logger stays at WARNING (clamp), not INFO. |
| 🟡 6 | Force-reinit keeps stale `module_levels` | Same snapshot mechanism: any logger previously managed but not in the new config is restored from snapshot before the new levels are applied. Test: `test_force_reinit_drops_module_level_no_longer_in_config`. |
| 🟢 7 | Level normalization duplicated | `_validate_log_level` now delegates to `normalize_log_level`. Single source of truth for `WARN→WARNING` + `VALID_LOG_LEVELS` membership. |

### Deferred (with reasoning — not bugs)

- **#8 — provider list duplicates registries.** Real architectural smell, but the fix is a multi-file refactor of `_VENDOR_PACKAGES` + `_BatchClientRegistration` to declare a `logger_root` per provider. That's a separate PR; doing it here would expand scope materially. The hardcoded list is now correct in the meantime.
- **#9 — denylist not config-extendable.** This is a feature add (YAML-extensible denylist), not a fix. Logged for a follow-up; the existing escape hatch (`module_levels`) handles the "want INFO from `httpx`" case which is the documented user need.
- **#10 — stderr helper.** Stylistic. Bootstrap-timing constraint is real (can't use `logger.warning` before the bridge is up); the 4 sites read fine inline. Would happily accept a follow-up PR.
- **#11 — level-name set duplicated 3×.** Real (and `level_map` has `WARN` while `VALID_LOG_LEVELS` doesn't — already drifted), but the divergence is acceptable: `level_map` is the only place that needs the `WARN` alias because it's the only place mapping user input to `EventLevel`. Worth a follow-up but not blocking.
- **#12 — guard altitude.** Two-encoding risk is real but limited: both sides reference `_root_logger_name`. Moving the bridge invariant into `_apply_logger_levels` would couple two concerns; current split is acceptable.

### Gates
- `ruff format --check`: 965 files clean
- `ruff check`: passed
- `mypy agent_actions`: 445 files clean
- `pytest`: 7682 passed (up from 7679 — 3 new regression tests)

PR #750, opened 2026-06-28
feat(VIOL-0008)!: consolidate compile/render into agac inspect

Muizzkolapo:
## Summary

Resolves VIOL-0008. `agac compile` and `agac render` are removed; the rendered-YAML output now lives behind `agac inspect --yaml`. `agac inspect` becomes the single preflight + introspection surface.

### Surface

```
agac inspect -a <wf>              # default: graph + validation status
agac inspect -a <wf> --yaml       # rendered YAML (replaces agac compile)
agac inspect -a <wf> --validate   # validation pass/fail
agac inspect -a <wf> --dry-run    # graph + validation + estimate
agac inspect -a <wf> --json       # JSON output (combines with above)

agac inspect graph    -a <wf>     # unchanged
agac inspect deps     -a <wf>     # unchanged
agac inspect action   -a <wf> N   # unchanged
agac inspect context  -a <wf> N   # unchanged
```

### Why

The old `inspect` commands instantiated the full `AgentWorkflow` (SQLite init, runtime services, event firing) just to read the action graph — wasted overhead on read-only commands. `agac compile` and `agac render` were duplicate code paths that bypassed `ConfigManager`, causing guard-terminology mismatches like `condition` vs `clause`.

### What changed

1. **`PreflightService`** — extracted from `coordinator._run_static_validation()` so runtime and CLI introspection share one validation implementation. The coordinator delegates and reuses the rebuilt `schema_service` (no double-build).
2. **`WorkflowInspector`** — lightweight read-only inspector. Drives the same `ConfigManager` pipeline as the runtime (so what inspect shows matches what `agac run` would execute) but skips storage, runtime services, and the `WorkflowInitializationStartEvent`. Validation runs with `verify_keys=False` so introspection never hits vendor endpoints.
3. **`InspectCommand` + group flags** — `agac inspect` is now a `click.group(invoke_without_command=True)` carrying the flag options. Subcommands still own their own options; the group banner is suppressed when routing to a subcommand to avoid double-print.
4. **`compile.py` deleted**, `main.py` no longer imports compile/render.
5. **17 new tests** covering all flag modes, mutual exclusion, JSON output, and regression guards that `compile`/`render` stay removed.
6. **Docs** — `inspect.md` rewritten, `utilities.md`/`index.md`/`templates.md`/`schemas/index.md`/`artifacts.md`/`output-validation.md`/SKILL.md updated to point at `agac inspect --yaml`.
7. **Review cleanup commit** — dropped dead `PreflightReport` dataclass, redundant `load()` call, misleading docstrings, and a test that registered signal handlers.

### Breaking changes

- `agac compile` and `agac render` are removed; CI/scripts using them must switch to `agac inspect --yaml`.
- The `-t/--template-dir` and `--create-dirs` flags have no equivalent under `agac inspect` (introspection is read-only).
- `agac inspect --yaml` output is the ConfigManager-merged YAML (correct), not the raw template dump the old `agac compile` produced. Documented in the migration table.

### Two intentional spec deviations

- The spec proposed keeping `_load_workflow` as a backward-compat alias for `_load_inspector`. RULES.md treats backward-compat shims as merge-blocking, so `_load_workflow` was deleted and all callers migrated.
- The spec's `@inspect.command(name="inspect", default=True)` is not valid Click. Used `invoke_without_command=True` on the group instead.

## Verification

- `pytest tests/ --ignore=tests/integration` → **7308 passed**
- `ruff check` / `ruff format` clean on all changed files
- Manual: `agac inspect --help` shows the new flag set; `agac compile` returns "No such command"
- Manual: `agac --help` no longer lists compile/render

## Commits

```
039c9005 refactor(VIOL-0008): review cleanup — drop dead state, fix doc claims
d2006598 docs(VIOL-0008): document unified agac inspect; retire compile/render
b4f47168 test(VIOL-0008): cover unified inspect flag matrix
6977aa1b feat(VIOL-0008)!: consolidate compile/render into agac inspect
a6fdafe3 refactor(VIOL-0008): extract WorkflowInspector for inspect commands
2cca181b refactor(VIOL-0008): extract PreflightService from coordinator
```


Muizzkolapo:
Review addressed in 87713eed. Per-item response:

### Substantive

1. **`create_project_paths(agent_name, agent_name)`** — verified the second positional is only used for error-message context (see `project_paths.py:228`). No path-resolution impact. Not changed.
2. **`resolve_project_root()` in coordinator** — verified `WorkflowResolutionService.self.project_root` is set but never read (only assignment in the file). No functional impact. The change also brings the resolution call into line with the schema-service call (which already passed `resolve_project_root()` even in the old code). Noted in the commit message.
3. **Docs `✅ Validation passed` line** — removed from `--dry-run` docs; replaced with a note that exit status carries the result.
4. **Inconsistent JSON shape** — unified on `{"status": "ok"}` everywhere. Default mode no longer emits `validation_ok`.
5. **`test_validate_json_fail` guarded assertion** — removed the `if result.output.strip()` guard; payload is asserted unconditionally now.

### Minor

6. **Group-flag positioning UX** — deferred (would need a custom Click option class to inspect the parent's value). Filed as follow-up.
7. **`Path(agent_opt).stem` duplication** — failure-path JSON now uses `cmd.agent_name` (which already computes the stem in `BaseInspectCommand.__init__`); the duplicated `Path(agent_opt).stem` is gone.
8. **`--user-code` untested** — turned out to be a real regression, not just untested: the flag was plumbed to `BaseInspectCommand.self.user_code` but never reached `WorkflowInspector`, so UDF discovery silently fell back to project `tool_path`. Added `user_code_path` to `WorkflowInspector.__init__`, threaded through `_load_inspector` and the `--yaml` fast path, with two tests asserting it reaches the inspector in both flows.
9. **Double-print in default mode** — restructured the group callback so non-JSON failures fall through to `@handles_user_errors` for a single uniform banner; JSON failures still emit JSON on stdout + `click.Exit(1)`. No more redundant "Error: Validation failed" line.
10. **`--yaml` skips validation** — added a `:::tip` block in `inspect.md` explaining the design and the workaround for users who want both.

### Cosmetic

11. **Triple-line stderr** — fixed by #9 above.
12. **`from rich.console import Console` unused** — verified there was no such import in `inspect.py` (only `from rich.tree import Tree`); `self.console` comes from `BaseInspectCommand`. No change.
13. **Always-true `validation_ok`** — addressed by #4 (replaced with `status`).

7310 tests pass; ruff clean. Ready for re-review.

Muizzkolapo:
Blockers addressed in 86dc3e1f:

1. **VIOL-0008 stripped from source/test** — removed from `cli/inspect.py` module docstring and `tests/unit/cli/test_inspect_consolidation.py` docstring. Ticket IDs now live only in commits/PRs/specs.
2. **Internal-symbol references cleaned across 5 files** — rewrote docstrings/comments in evergreen terms:
   - `workflow_inspector.py`: module / class / `render()` / `load()` docstrings no longer mention `BaseInspectCommand._load_workflow`, `AgentWorkflow`, `load_workflow_configs`, `agac compile`, `agac render`.
   - `preflight_service.py`: module + class docstrings no longer say "Extracted from `AgentWorkflow._run_static_validation()`". The lazy-import comment no longer enumerates `coordinator`'s transitive deps.
   - `coordinator.py`: `_run_static_validation` docstring no longer narrates the delegation to `PreflightService`.
3. **Docs/code schema mismatch fixed** — `inspect.md` now shows `"status": "ok"` (matches code) instead of `"validation_ok": true`, with a one-line note that `status` is the consistent key across `--validate`, `--dry-run`, and default JSON output.

All inspect tests pass; ruff clean.

Muizzkolapo:
Round-3 review addressed in c8cf78e3. Per-item:

### 🔴 Critical

**1. `--yaml` always emitted `actions: []` for new-format configs** — CONFIRMED. `ConfigRenderingService` writes parsed actions to `_validated_actions`, not under `agent_name`. Fixed `render()` to read `_validated_actions` first, fall back to `agent_name` for legacy format. The previous tests mocked `render()` entirely so this path wasn't exercised — added two direct `WorkflowInspector.render()` tests (new format + legacy format) that would catch regressions.

### 🟠 Contract break

**2. `--yaml --json` silently ignored `--json`** — CONFIRMED. Added explicit mutex (`--yaml` emits raw YAML and cannot be wrapped in JSON). Added regression test. Docs updated.

### 🟡 Plausible

**3. Coordinator `resolve_project_root()` change** — REFUTED (third time). `WorkflowResolutionService.self.project_root` is set at line 110 but never read — only `find_project_root_dir(config_start)` is used (line 546, a local computation independent of the constructor arg). The change has zero functional impact. Happy to discuss if you have a specific call path in mind.

**4. `--yaml` doesn't reflect `apply_guard_nullable_schema_fixes()`** — CONFIRMED docs accuracy issue. The transform happens in `validate()`, not `render()`. Rewrote the docs `:::tip` block to be explicit: `--yaml` reflects the *rendered* configuration; runtime-only transforms (including guard-nullable fixes) are not applied. Users wanting that view should use `--validate` or `--dry-run`.

**5. `Console(quiet=True)` swallowed UDF warnings** — PARTIALLY REFUTED, partially CONFIRMED. Warnings actually go through `logger.warning(...)` (which `quiet=True` does not touch); only the success/info banners route through the console. But the UX gap was real — users had no visibility into UDF discovery during inspect. Switched to `Console(stderr=True)` so banners appear without polluting `--json` stdout.

**6. `get_levels()` drops non-operational agents and can fake a cycle** — CONFIRMED. `ConfigManager.determine_execution_order()` only ranks `is_operational` agents (line 321). The old code filtered against `execution_order` and could put non-operational agents in a "cycle" bucket. Fixed: walk `execution_order` first, then append any `action_configs` entries it missed. Added a test asserting non-operational actions land in correct topological position after their deps, not in a cycle bucket.

7314 tests pass; ruff clean.

PR #753, opened 2026-07-03
fix(VIOL-0009): FILE-mode empty output honors on_empty config

Muizzkolapo:
## Summary

`FileToolStrategy` ignored the workflow config's `on_empty` field. The schema at `agent_actions/config/schema.py:231` declares `on_empty: Literal[\"warn\", \"error\", \"skip\"]` on every action, and `online_llm.py:512-560` honors all three — but `file_tool.py:76-88` unconditionally returned per-record `ProcessingResult.failed(...)`, silently collapsing `error` and `skip` to the `warn` semantic. A workflow declaring `on_empty: error` on a FILE-mode tool did **not** abort; declaring `on_empty: skip` did **not** produce `empty_output` tombstones.

This PR brings FILE-mode to parity by adding the same three-way dispatch, reusing the existing `EmptyOutputError` and `EMPTY_OUTPUT` reason string. The `warn` branch is byte-identical to the prior behavior, so no existing workflow regresses.

## Root cause

`context.agent_config` was available (used at line 66 for cast) but never queried for `on_empty` on the empty-response branch. The fix reads it and branches like `online_llm.py` does.

## Verification

- RED commit `eee85faf` adds `tests/unit/workflow/test_file_tool_on_empty_parity.py` with 4 tests; the two parity tests (`test_on_empty_error_raises`, `test_on_empty_skip_produces_empty_output_tombstones`) fail on that commit for the right reason (\"DID NOT RAISE\"; status FAILED instead of SKIPPED). Baseline `warn` and default-unset tests pass.
- GREEN commit `d04fd25e` applies the fix — all 4 parity tests pass.
- Full FILE-mode + processing suites: **486 passed** (`pytest tests/unit/processing/ tests/unit/workflow/test_pipeline_file_mode_tool.py tests/unit/workflow/test_file_tool_on_empty_parity.py`).
- `ruff format --check` and `ruff check` clean on changed files.
- 28 pre-existing failures elsewhere (`test_parallel_exception_handling.py`, `test_stale_completion_verification.py`) are on `main` too — unrelated.

## Why not add a new `fail_fast_on_empty_file_mode_output` setting?

An earlier spec draft proposed a parallel setting and a new exception class. Both would duplicate existing mechanisms (`on_empty=\"error\"` and `EmptyOutputError`). The real defect is that FILE-mode wasn't wired to the existing field — not a missing field.

### DIFF
--- a/agent_actions/processing/strategies/file_tool.py
+++ b/agent_actions/processing/strategies/file_tool.py
@@ -7,13 +7,14 @@ import logging
 from typing import Any, cast
 
 from agent_actions.errors import AgentActionsError
+from agent_actions.errors.processing import EmptyOutputError
 from agent_actions.processing.helpers import run_dynamic_agent
 from agent_actions.processing.record_helpers import build_tombstone
 from agent_actions.processing.types import (
     ProcessingContext,
     ProcessingResult,
 )
-from agent_actions.record.reasons import TOOL_MISSING_RECORD
+from agent_actions.record.reasons import EMPTY_OUTPUT, TOOL_MISSING_RECORD
 from agent_actions.record.tracking import TrackedItem
 from agent_actions.utils.content import is_version_merge
 from agent_actions.utils.tools_resolver import resolve_tools_path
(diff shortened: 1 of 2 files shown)


Muizzkolapo:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [pi](https://github.com/earendil-works/pi)

PR #763, opened 2026-07-05
fix(VIOL-0034): cascade-skip instead of crashing when all version branches filtered

Muizzkolapo:
## Summary
- A `version_consumption: merge` action whose every version branch was guard-filtered **crashed the pipeline** with `ConfigurationError: Version correlation failed` (exit 1). All-branches-filtered is a legitimate low-density input shape, not a misconfiguration. Root cause: `resolve_correlated_input` (`workflow/managers/output.py`) raised unconditionally whenever correlation returned nothing, and the executor called it **uncaught** (`_execute_action_run`, `executor.py:1145`, outside the try/except).
- `resolve_correlated_input` now distinguishes **all-sources-empty** (every version source has no target files) → raises the typed `AllVersionsFilteredError` — from a **genuine correlation failure** (some source had output) → still raises `ConfigurationError`.
- The executor catches `AllVersionsFilteredError` at both action-run sites (sync + async) and resolves the action as **SKIPPED** (`reason=all_versions_filtered`, node-level disposition), mirroring the existing `_handle_all_guard_filtered` "produced nothing → SKIPPED" pattern — so the pipeline exits 0. Downstream consumers then cascade-skip via the normal mechanism.

## Verification
- New `tests/unit/workflow/managers/test_resolve_correlated_input_all_filtered.py`, RED→GREEN:
  - all sources empty → `AllVersionsFilteredError`
  - some source has output → `ConfigurationError` (genuine failure preserved)
  - executor `_execute_action_run` cascade-skips (SKIPPED, no raise) instead of propagating — pins the actual crash path, not just the isolated raise
- Existing `test_version_correlation_failure_raises_error` comment corrected: its mock storage reports outputs, so it exercises the genuine-error `ConfigurationError` path; still green.
- Blast radius: `tests/unit/workflow/` + `tests/core/graph/` → **826 passed**. `ruff check`/`format` clean; `uv run mypy agent_actions` → no issues (447 files).

### DIFF
--- a/agent_actions/record/reasons.py
+++ b/agent_actions/record/reasons.py
@@ -19,6 +19,7 @@ LLM_LAYER_GUARD_FILTER = "llm_layer_guard_filter"
 # -- Cascade / upstream ------------------------------------------------------
 UPSTREAM_UNPROCESSED = "upstream_unprocessed"
 OBSERVE_FIELD_MISSING = "observe_field_missing"
+ALL_VERSIONS_FILTERED = "all_versions_filtered"
 
 # -- Prep failures -----------------------------------------------------------
 PREP_FAILED = "prep_failed"
--- a/agent_actions/workflow/executor.py
+++ b/agent_actions/workflow/executor.py
@@ -20,7 +20,7 @@ from agent_actions.logging.events import (
     BatchCompleteEvent,
     BatchSubmittedEvent,
 )
-from agent_actions.record.reasons import GUARD_FILTERED_ALL
+from agent_actions.record.reasons import ALL_VERSIONS_FILTERED, GUARD_FILTERED_ALL
 from agent_actions.storage.backend import (
     DISPOSITION_FAILED,
     DISPOSITION_FILTERED,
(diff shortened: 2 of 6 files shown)


Muizzkolapo:
### Code review

Found 1 issue:

1. `_handle_all_versions_filtered` writes the node-level SKIPPED disposition through an unguarded `storage_backend.set_disposition(...)`, bypassing the `_write_skipped_disposition` helper (`executor.py:686`) that every other skip path uses (e.g. `_skip_action_upstream_failure`). That helper wraps the identical call in `try/except Exception ... logger.warning(...)`; here a storage write failure (e.g. a SQLite lock) would propagate out of `_handle_all_versions_filtered` and out of `_execute_action_run`, crashing the very path meant to cascade-skip gracefully. Route through `_write_skipped_disposition` (extended with an optional `detail=`) or wrap the call in the same try/except.

https://github.com/Muizzkolapo/agent-actions/blob/bc39aa0887f18e4df9eaa25e984136d3c0010d61/agent_actions/workflow/executor.py#L573-L581

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #770, opened 2026-07-06
fix(VIOL-0034): cascade-skip version-merge by cause, not file existence

Muizzkolapo:
## Summary

Complete, corrected fix for the version-merge all-branches-filtered crash. Supersedes #763 — it contains #763's cascade-skip scaffolding plus the fixes for the correctness gaps a follow-up review found, so **#763 can be closed in favor of this**.

The core defect in #763's approach: `resolve_correlated_input` decided "all sources empty" from `list_target_files` (file existence). That mis-scopes both directions.

### Fixed
- **Empty-output-file variant still crashed.** A guard-filtered branch that wrote an empty output file (`[]`) has files but zero records, so it was read as "has output" and fell through to a `ConfigurationError` the executor does not catch — re-raising the very crash this path prevents.
- **Silent masking / exit 0.** Any all-empty state cascade-skipped and exited 0, hiding sources that completed with zero output for non-filter reasons (missing data) — which previously raised.
- **Unguarded storage probe.** The discriminator's storage reads were unguarded; a backend error escaped as a raw `sqlite3.Error`/`OSError` instead of a clean, actionable error.
- **Unguarded disposition write on the skip path** (carried from #763's follow-up): the node-level skip disposition now routes through the best-effort helper, so a storage hiccup can't re-crash the cascade-skip.

### Approach
Classify by **cause**, not file existence:
- a source with **output records** → correlation genuinely failed → `ConfigurationError`;
- **no records but guard-filtered** → `AllVersionsFilteredError` → executor cascade-skips;
- **no records and not filtered** → missing data → `ConfigurationError` naming the sources.

Record counts come from the already-guarded backend loader; the filtered-disposition probe (`has_disposition(src, filtered)`) is guarded — a storage error surfaces a clear `ConfigurationError`, never a raw crash.

## Verification
- New/updated tests: empty-output-file skip (the crash variant), has-output failure, all-empty-not-filtered failure, guarded disposition-probe error; sync + async raising-backend skip paths; node-level SKIPPED disposition assertion. Correlator failure test now configures explicit output records instead of relying on incidental mock truthiness.
- Proved non-tautological: reverting the record-count signal to file existence makes the empty-output-file test fail with the original crash; reverting the disposition-write guard makes the raising-backend tests fail.
- `pytest` managers + correlator + executor suites: 263 passed. `ruff check`/`format --check` clean. `mypy` clean.
- Manifest updated to document the public `AllVersionsFilteredError` and its catch contract.

## Follow-ups (lower severity, not in this PR)
- Redundant per-source re-query in the discriminator (have the correlator return per-source counts).
- `ALL_VERSIONS_FILTERED` token vs `GUARD_FILTERED_ALL` sentence — skip-reason taxonomy consistency.
- Shared `_resolve_as_skipped(...)` extraction between the two skip handlers.


Muizzkolapo:
Superseded by a clean branch off updated main (VIOL-0034 #763 now merged). Reopening the same fix as a fresh PR.

PR #779, opened 2026-07-10
fix(inspect): fail preflight when a dependency has no observe field

Muizzkolapo:
## Summary
- `agac inspect` (and `agac run` preflight) now hard-fails when a declared dependency has no `observe`/`passthrough` field reference in `context_scope` — previously inspect reported green and the run crashed on the first record inside `scope_namespace._extract_allowed_fields_per_dependency`, after upstream LLM tokens were already spent (spec 565 / ISSUE-007).
- New pure checker `agent_actions/validation/dep_observe_validator.py` (`find_missing_observe_deps`): mirrors the fatal runtime check but fires no events and reports **all** offenders instead of raising on the first. Wired into `PreflightService.validate()` after guard-syntax validation, so both the inspect path (`workflow_inspector.py`) and the run path (`coordinator.py`) are covered. The runtime check stays as the last line of defence.
- Namespace parity is deliberate: refs are parsed with the runtime's own `parse_field_reference`, skipping refs that raise — a dotless ref like `observe: ["producer"]` does not satisfy a dependency here because it does not satisfy it at execution time. A naive `split('.')` (the spec's original draft) diverges exactly there and was rejected during spec verification.
- Docs: `reference/cli/inspect.md` preflight coverage sentence extended; validation `_MANIFEST.md` row added.

## Verification
- Red→green: behavioral tests committed failing first (65852092) — `PreflightService.validate()` DID NOT RAISE on an over-declared dep, and end-to-end `agac inspect` on a real scaffolded project exited 0. Both flip with the wiring commit.
- Live probes on a scaffolded project (pre-fix: inspect exit 0 "validated" for both shapes; runtime `ConfigurationError` confirmed for both): post-fix inspect exits 1 naming the exact dependency, message parity with the runtime error.
- Unit tests include discriminators that each kill a specific wrong implementation: dotless ref (naive split), substring dep name (`dep in ref`), drop-only scope, empty-scope dict, all-offenders count (raise-on-first).
- Full suite: **7921 passed**; `ruff check` + `ruff format --check` clean; no `ISSUE-`/`VIOL-` identifiers in source or tests.
- Known residual (not a blocker, cannot false-positive): versioned/fan-in dep expansion is not mirrored — the check reads the loader-validated `dependencies: list[str]` directly, so it can only under-flag those shapes; the runtime check still catches them.

### DIFF
--- a/agent_actions/services/preflight_service.py
+++ b/agent_actions/services/preflight_service.py
@@ -12,6 +12,7 @@ from agent_actions.errors.preflight import PreFlightValidationError
 from agent_actions.models.action_schema import FieldSource
 from agent_actions.prompt.formatter import PromptFormatter
 from agent_actions.utils.constants import NON_PROMPT_ACTION_KINDS
+from agent_actions.validation.dep_observe_validator import find_missing_observe_deps
 from agent_actions.validation.preflight.guard_validation import validate_guard_conditions
 from agent_actions.validation.preflight.resolution_service import (
     WorkflowResolutionService,
@@ -81,7 +82,16 @@ class PreflightService:
                 hint="Fix the guard condition errors above before running the workflow.",
             )
 
-        # 4. Resolution checks (API keys, seed files, vendor batch)
+        # 4. Dependency/observe consistency — fatal at runtime, so fatal here
+        observe_errors = find_missing_observe_deps(self.action_configs)
+        if observe_errors:
+            raise PreFlightValidationError(
+                "\n".join(observe_errors),
+                hint="Add an observe or passthrough field for each dependency, "
+                "or drop the unused dependency.",
+            )
+
+        # 5. Resolution checks (API keys, seed files, vendor batch)
         resolution_result = WorkflowResolutionService(
             action_configs=self.action_configs,
             workflow_config_path=self.workflow_config_path,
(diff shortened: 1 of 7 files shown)


Muizzkolapo:
**Second verification round — real-project smoke on qanalabs-quiz-maker found and fixed a false positive.**

Running `agac inspect` across all 10 real workflows exposed a divergence the scaffolded probes and 7921-test suite missed: the checker read `cfg["dependencies"]` raw, but for version-consumption actions the loader expands the producer into `<base>_N` branches and rewrites the consumer's `observe` refs to the branch names while `dependencies` keeps the base name. Result: `qana_quiz` (4 findings) and the versioned half of `code_centered_quiz` were hard-failed at inspect despite the runtime accepting them (verified by running `infer_dependencies` + `_extract_allowed_fields_per_dependency` directly on the loaded configs).

Fix (`32826ead`, red test first in `693e9813`): the checker now derives its dependency set via `infer_dependencies(validate=False)` — the runtime's own derivation in `scope_builder` — instead of raw config strings. This removes the whole divergence class (string form, fan-in repartition, template inference, version expansion), not just the observed shape.

After the fix, the real-project sweep reads exactly as it should:
- 9/10 workflows pass inspect.
- `code_centered_quiz` fails naming `review_quiz_before_assembly → reconcile_distractor_explanations` — a **true positive**: the runtime check raises on identical inputs, i.e. that action crashes at execution today. The check caught a real latent bug in a production workflow at inspect time.

Full suite 7925 pass, ruff clean. New regression coverage at all three layers: unit (expanded branches satisfy a base-name dep + unreferenced branches still flagged), preflight (post-loader versioned shape), and end-to-end (real fan-out/merge project through the CLI loader).

Note for reviewers: the earlier claim (in the spec's second-pass review and this PR's initial self-review) that the versioned/fan-in residual "can only under-flag" was wrong — this was an over-flag. The lesson is recorded in the spec.

PR #781, opened 2026-07-11
feat(validate-udfs): warn on UDF reads of unknown bus namespaces

Muizzkolapo:
## Summary

The UDF `data` bus is keyed by **action name**. `data.get("wrong_name")` returns `{}` silently — a namespace typo (e.g. reading the *impl* name instead of the *action* name) yields an empty dict with no error, so downstream gates reject every record while the run reports success (`36/36 ✓`). Diagnosis today means poking the SQLite store by hand.

This adds a static preflight to `agac validate-udfs` / `inspect`:

- **`agent_actions/validation/bus_namespace_validator.py`** — `find_unknown_bus_namespaces()` AST-scans each UDF source for literal `data.get("X")` / `data["X"]` reads and warns when `X` is outside the valid namespace set. Dynamic keys (`data.get(var)`), non-`data` variables, and literals inside comments/strings are ignored. Each finding is attributed to its **enclosing UDF** via a function-name stack, so one file holding several UDFs never cross-attributes.
- **`agent_actions/validation/validate_udfs.py`** — `validate()` now returns the workflow's `action_names`; `execute()` unions them with `SPECIAL_NAMESPACES` and prints one `⚠` per unknown read. It stays a **warning, not an error** (exit 0) so intentional optional reads (`data.get("optional", default)`) stay unbroken.

## Deviations from the spec (called out)

1. **Valid framework set = `SPECIAL_NAMESPACES`, not `{"seed","source"}`.** The spec hardcoded 2 framework namespaces; the authoritative reserved set (`agent_actions/utils/constants.py`) has 7 (`action, prompt, schema, seed, source, version, workflow`). Using the full set removes false positives on legitimate reads like `data.get("workflow")`; reserved names can't be action names, so there is no false-negative risk. The bus keying (by action name) was confirmed against `schema_output_validator.py`.
2. **Attribution to the enclosing function + `sources` keyed by file path.** The registry records one `file` per function, so the spec's `{name: read_text(file)}` map would scan a shared file twice and blame both functions for one typo. Keying `sources` by file (dedup) plus enclosing-function attribution fixes this.
3. **Commit structure.** A behavioral (command-level) RED lands first — the harness RED gate rejects `ImportError` collection errors as fake-RED — then GREEN adds the module, the wiring, and the pure-function unit tests together.

## Verification

- `pytest tests/validation/test_bus_namespace_validator.py tests/validation/test_validate_udfs.py` — pure-function unit tests (unknown flagged, known/dynamic/non-`data`/comment-string ignored, enclosing-function attribution, dedup, syntax-error skip, default-arg still flagged) + command-level tests (typo warns + exit 0; action name & framework namespace not flagged).
- Full suite: **7942 passed**; `ruff check` and `ruff format --check` clean.
- Blast radius: extended `validate()`'s return contract with `action_names`; updated the only other consumer mocks (`test_validate_udfs_exit_code.py`, `TestExecute`). No other callers of the return.
- **Live end-to-end** (`agac validate-udfs -a contract_reviewer -u tools` on `examples/contract_reviewer`): a UDF reading `data.get("split_into_clause")` (typo for action `split_into_clauses`) produced `⚠ probe_typo: reads bus namespace 'split_into_clause' ...`, while `data.get("analyze_clause")` (real action) and `data["seed"]` (framework namespace) were **not** flagged; exit code **0**.
- Blind fresh-context review: **YES** (correct AST scan, valid-namespace set, exit-0, dedup/attribution; no missed siblings).

**Smoke:** skipped — `risk.sh` reported `smoke_required=no` (validation-only, no chokepoint surface). The live `contract_reviewer` run above already exercised the real CLI path.

### DIFF
--- a/agent_actions/utils/constants.py
+++ b/agent_actions/utils/constants.py
@@ -100,6 +100,15 @@ SCHEMA_FILE_GLOBS = tuple(f"*{s}" for s in SCHEMA_SUFFIXES)
 # - Context scope processor
 SPECIAL_NAMESPACES = RESERVED_AGENT_NAMES - {"context_scope"}
 
+# Framework-injected keys that the runtime actually writes into a tool UDF's
+# `data` bus, alongside per-dependency action-name namespaces. Populated by
+# build_field_context_with_history (source/version/workflow) + scope_application
+# (seed). Deliberately narrower than SPECIAL_NAMESPACES: prompt/schema/action are
+# reserved config/agent-name tokens that never appear as runtime bus keys, and
+# "loop" is the Jinja loop variable, not bus data. Version-promoted keys
+# (i/idx/custom params) are dynamic and not statically enumerable here.
+RUNTIME_BUS_NAMESPACES = frozenset({"source", "version", "workflow", "seed"})
+
 HITL_FILE_GRANULARITY_ERROR = (
     "HITL actions require FILE granularity. "
     "Record granularity launches a separate approval UI per record. "
--- a/agent_actions/validation/_MANIFEST.md
+++ b/agent_actions/validation/_MANIFEST.md
@@ -23,6 +23,7 @@ decoders to schema validators and preflight checks.
 |------|------|-------------|---------|
 | `base_validator.py` | Module | `BaseValidator` base class with helper assertions for validators. | `validation` |
 | `batch_validator.py` | Module | Validator that ensures batch CLI arguments conform to expectations. | `validation`, `llm.batch` |
+| `bus_namespace_validator.py` | Module | `find_unknown_bus_namespaces`: AST scan flagging tool-UDF reads of `data.get("X")` / `data["X"]` where X is not a runtime bus namespace (action name or framework key), catching silent namespace typos at `validate-udfs`. | `validation` |
 | `clean_validator.py` | Module | `CleanCommandArgs` pydantic model used by the CLI. | `validation` |
 | `config_validator.py` | Module | Central config parser/validator used across startup flows. | `configuration`, `validation` |
 | `dep_observe_validator.py` | Module | `find_missing_observe_deps`: preflight mirror of the fatal runtime check that every declared dependency has an observe/passthrough field reference. | `validation` |
--- a/agent_actions/validation/bus_namespace_validator.py
+++ b/agent_actions/validation/bus_namespace_validator.py
@@ -1,0 +1,89 @@
+"""Flag UDF reads of `data.get("X")` / `data["X"]` where X is not a bus namespace."""
+
+from __future__ import annotations
+
+import ast
+
+
+def _bus_key(node: ast.AST) -> str | None:
+    """Return the literal key of a `data.get("X")` / `data["X"]` read, else None."""
+    # data.get("literal")
+    if (
+        isinstance(node, ast.Call)
+        and isinstance(node.func, ast.Attribute)
+        and node.func.attr == "get"
+        and isinstance(node.func.value, ast.Name)
+        and node.func.value.id == "data"
+        and node.args
+        and isinstance(node.args[0], ast.Constant)
+        and isinstance(node.args[0].value, str)
+    ):
+        return node.args[0].value
+    # data["literal"]
+    if (
+        isinstance(node, ast.Subscript)
+        and isinstance(node.value, ast.Name)
+        and node.value.id == "data"
+        and isinstance(node.slice, ast.Constant)
+        and isinstance(node.slice.value, str)
+    ):
+        return node.slice.value
+    return None
+
+
+class _BusKeyCollector(ast.NodeVisitor):
+    """Collect (enclosing-function, literal bus key) pairs; innermost function wins."""
+
+    def __init__(self, fallback: str) -> None:
+        self._fallback = fallback
+        self._func_stack: list[str] = []
+        self.reads: list[tuple[str, str]] = []
+
+    def visit_FunctionDef(self, node: ast.FunctionDef | ast.AsyncFunctionDef) -> None:
+        self._func_stack.append(node.name)
+        self.generic_visit(node)
+        self._func_stack.pop()
+
+    visit_AsyncFunctionDef = visit_FunctionDef
+
+    def _record(self, node: ast.AST) -> None:
+        key = _bus_key(node)
+        if key is not None:
+            owner = self._func_stack[-1] if self._func_stack else self._fallback
+            self.reads.append((owner, key))
+
+    def visit_Call(self, node: ast.Call) -> None:
+        self._record(node)
+        self.generic_visit(node)
+
+    def visit_Subscript(self, node: ast.Subscript) -> None:
+        self._record(node)
+        self.generic_visit(node)
+
+
+def find_unknown_bus_namespaces(sources: dict[str, str], valid_namespaces: set[str]) -> list[str]:
+    """Return one warning per literal bus key that is not a valid namespace.
+
+    Findings are attributed to the enclosing UDF, so a source file holding several
+    UDFs never cross-attributes; a `sources` key is only the fallback label for a
+    read outside any function. Non-literal keys (`data.get(var)`) are skipped.
+    """
+    findings: list[str] = []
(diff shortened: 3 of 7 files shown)


Muizzkolapo:
## Update — real-project run exposed a false-positive flood; fixed (commit `fd044b25`)

After opening this PR I ran `agac validate-udfs` against the real **qanalabs-quiz-maker** project (a shared `tools/` dir: 67 discovered UDFs, 23 referenced per workflow). It produced **33 false-positive warnings**:

- **Registry-wide scan.** The scan walked the *entire* discovered registry, so every *other* workflow's tool got checked against *this* workflow's action names. (Same root cause the FILE-mode contract warning hit — it was fixed there by scoping to `impl_refs`.)
- **`data["X"]` is statically ambiguous.** Record-level helpers (e.g. `process_single_mcq` reading `data.get("question")`, `data.get("options")`) and the `content = data.get("content", data)` wrapper-unwrap pattern are legitimate non-bus reads that got flagged.

### Fix
Scope the scan to `impl_refs` **and** to each referenced UDF's *own function body* via `inspect.getsource` (not its whole file), so a shared file's helper is not cross-attributed.

### Verified on qana
- Both workflows now warn **0** times (down from 33).
- Injected a real typo into a *referenced* UDF — `select_fields` reading `data.get("flatten_canonical_question")` (action is `flatten_canonical_questions`) — and it still warns: `⚠ select_fields: reads bus namespace 'flatten_canonical_question' ...`, exit 0. So scoping did not over-suppress.
- Full suite **7944 pass**; ruff clean. RED (a new `test_unreferenced_udf_in_shared_dir_is_not_scanned` that fails on the registry-wide scan) → GREEN. Two scoping discrimination tests added. Fresh blind re-review: **YES**.

### Known limitation (documented, not code-fixed)
`data["X"]` cannot be statically disambiguated (bus-read vs record-field vs `content`-wrapper). Scoping removes every false positive observed on the real project, but a *referenced* record-granularity UDF that reads a field via `data["field"]` could still warn. The robust long-term fix is the runtime `bus.get()` accessor the spec notes as out of scope.

Muizzkolapo:
## Response to the review (verified each finding against current HEAD; fixes in `697d5678`)

Heads-up first: several findings cite `Path(meta["file"]).read_text(...)` at `validate_udfs.py:190` and "whole registry scanned" — that code was replaced by commit `fd044b25` (scan scoped to `impl_refs`, each UDF read via `inspect.getsource`). So the review appears to have run against the pre-scoping revision `43f26f03`. I re-verified every finding against current HEAD; verdicts below.

### Fixed
- **#1(a) — namespace set was wrong (confirmed).** Traced the runtime bus end-to-end (`ToolClient.invoke` → `prepare_context_data(is_tool=True)` → `build_field_context_with_history`): the top-level keys are action names + `{source, version, workflow, seed}`. `prompt`/`schema`/`action` are reserved config/agent-name tokens that are **never** bus keys, so whitelisting them (via `SPECIAL_NAMESPACES`) silently let those typos through. Added `RUNTIME_BUS_NAMESPACES = {source, version, workflow, seed}` (`constants.py`) and switched the validator to `action_names | RUNTIME_BUS_NAMESPACES`. `data.get("schema")` now warns.
- **#4 — Rich markup injection (confirmed, live repro).** A key like `[/x]` raised `MarkupError` (crash) or silently dropped the key. Now `escape()`d before printing.
- **dedup (#2, PARTIAL) — confirmed narrow case.** Top-level collisions are impossible (unique registered names), but two UDFs with identically-named *nested* helpers reading the same key collapsed to one finding via a global `seen`. Dedup is now per-source.
- **#11 — test isolation was illusory (confirmed, and worse).** `clear_registry`/`clear_module_cache` do **not** evict `sys.modules['agent_actions._udfs.*']`, so a reused file stem returned an empty module — and because the scoping regression tests only asserted negatives, they passed **vacuously**. Fixed the fixture to evict those entries (the established `_evict_udf_modules` pattern) and added positive assertions so the tests fail if the referenced UDF isn't scanned.
- **#15 — manifest.** Added the `bus_namespace_validator.py` row to `validation/_MANIFEST.md`.

### Refuted (with evidence)
- **#1(b) — "`loop` is a real bus key" → REFUTED.** `loop` is the **Jinja2 loop variable** (`{{ loop.index }}`), documented as such (`reference_extractor.py:29`, `scope_parsing.py:158`). No runtime code writes a `"loop"` key into the bus; iteration data lives under the `version` namespace. Adding `loop` would *suppress* a genuine typo. `FRAMEWORK_NAMESPACES` over-includes it for Jinja-render gating, which is why it's not the right source of truth.
- **#7 — data-reassignment false positive.** The cited example (`aggregate_quality_scores.py`) does **not** reassign `data`; it uses a different variable (`scorer_ns`) and a **dynamic** key (`data.get(scorer_key)`), which the scanner already skips. The `data = data.get("x")` drill-down pattern is a real (rare) residual, folded into the documented `data["X"]`-ambiguity limitation below.
- **#14 — reuse `code_scanner.scan_tool_functions`.** Different abstraction: it's a path-based project indexer (`scan_tool_functions(project_root, tool_paths) -> dict`) returning tool functions; this validator takes in-memory source strings and pattern-matches bus reads with enclosing-function attribution. Forcing reuse would add coupling without reducing real complexity.

### Moot on current HEAD (targeted the pre-scoping code; re-verified)
- **#6 (helper with `data` param), #13 (double disk read):** MOOT — `getsource` scans only the referenced UDF's own function; no `read_text`.
- **#5 (BOM), #9 (latin-1):** MOOT for the scan — `getsource` reads via tokenize/linecache which honor BOM and the PEP-263 cookie. (Separately: `_declares_tool_decorator` in `input/loaders/udf.py:22` does hard `read_text(encoding="utf-8")` and silently skips a latin-1 tool file at **discovery** — a real but out-of-scope discovery-layer bug; filing separately.)
- **#8 (decorator-arg attribution):** theoretical only — a real `@udf_tool(...)` decorator's args never call `data.get(...)`.

### Known/accepted (documented limitations, not regressions)
- **#3 — `agac inspect` doesn't run this.** Correct: `inspect` uses `use_tools=False` (`workflow_inspector.py:80`), so it never discovers UDFs. This is validate-udfs-only, the same tradeoff as the sibling FILE-mode contract check; wiring inspect needs UDF discovery there (larger change), deferred.
- **#10 (`self.data`), version-promoted keys (`i`/`idx`/custom params), and FILE-mode field-flattened dicts:** all inherent limitations of a coarse static matcher on the ambiguous `data["X"]` read. Warning-only by design; the robust fix is the runtime `bus.get()` accessor noted as out of scope.
- **#12 (no strict mode):** warning-only is the intended severity (a strict lever is speculative/YAGNI until requested).

### Verification
Full suite **7949 pass**, ruff clean. Re-verified live on `qanalabs-quiz-maker`: both workflows warn **0** times with the tightened set (no new false positives from dropping `prompt`/`schema`/`action`). Three diverse-lens blind reviewers (correctness / adversarial / test-quality) all returned YES on the updated diff.

(Housekeeping: feel free to delete the local `pr-781-review` branch and `/tmp/pr781-worktree` scratch state from the review run — nothing there is needed.)

Muizzkolapo:
## Merged `main` — conflicts resolved (`01a489e5`)

`main` advanced by #780 (the sibling FILE-mode contract warning), which touched the same `execute()` region and appended a test class to `test_validate_udfs.py`. Resolved:

- **`validate_udfs.py` `execute()`**: both preflights now run and print together. Consolidated into a single warning list + one print loop, and `escape()` is applied to **both** warning sources — which also hardens #780's contract warning against the same Rich-markup class (a `list[dict]` return annotation contains `[dict]`).
- **`test_validate_udfs.py`**: kept both test classes (`TestBusNamespaceWarnings` + `TestFileUdfContractWarnings`). Added `"action_names": set()` to #780's three `validate()` mocks, since the merged `execute()` now reads `result["action_names"]`.

Verification: full suite **7963 pass**, ruff clean. Live on qanalabs-quiz-maker: bus-namespace check still **0** warnings; the one warning now shown is #780's legitimate FILE-mode warning on `convert_html_json_to_thinkific` (exit 0) — both features work independently, no cross-interference.

PR #784, opened 2026-07-12
fix(udf-discovery): frame auto-discovered import failures with file and next step

Muizzkolapo:
## Summary
When a `.py` file under the `-u` user-code tree is auto-discovered by the recursive UDF scan and fails to import, the CLI error title read `Failed to load UDF module 'scripts.fix_prompts'` — as if the user had named that module. They didn't; the recursive walk found it. The title now names the offending **file**, states it was **auto-discovered**, keeps the underlying import error visible, and gives an actionable next step. Spec: `specs/active/570-udf-discovery-error-framing.md` (ISSUE-006 rec #3; follow-up to 564).

Before → after (rendered via `format_user_error`):
```
- Configuration Error: Failed to load UDF module 'scripts.fix_prompts'
+ Configuration Error: Auto-discovered UDF file failed to import: tools/scripts/fix_prompts.py

  Problem: Python could not import the UDF module: [Errno 2] No such file or directory
  File: tools/scripts/fix_prompts.py
  Fix: ... This file was found by the recursive scan of the -u tree. Fix its
       module-load code, or — if it is not a UDF — move it outside the -u
       directory so discovery skips it. ...
```

Scope: `_format_import_failure` (non-sentinel branch) only — `title` reframe + the `-u`-tree next step. `details` (the cause), the ModuleNotFoundError install guidance, and the directory-level discovery-sentinel branch (`UDF discovery failed`) are all unchanged. The fix wording also drops a now-stale claim — post-564, discovery no longer "imports every Python file," so "found by the recursive scan of the -u tree" is the accurate framing.

## Deviations from the spec
- **Spec's test file/path were wrong.** The spec pointed at `tests/logging/errors/formatters/test_udf_load_formatter.py` (does not exist) and `UDFLoadErrorFormatter().format(err)` (wrong signature — `format` takes `(exc, root, message, context)` and returns a `UserError`, not a string). Real test file: `tests/unit/error_formatters/test_udf_load.py`. Followed the spec's appended adversarial review, which had already corrected these.
- **Blast radius was larger than documented.** The adversarial review named four old-title assertions (lines 69, 310, 343, 368). Verify-first found **two more**: `test_udf_load.py:227` (`format_for_cli` prefix) and `tests/validation/test_validate_udfs.py:311` (the `validate-udfs` command renders the same formatter). Six title assertions updated in total, across two files.
- **`configuration.py:114` `str(exc)` left as-is (deliberate).** `UDFLoadError.__init__` builds a separate `Failed to load UDF module 'X'` string for `str(exc)`. It is log-only and never user-facing — the blind reviewer confirmed `config_pipeline.py:165`'s warning uses `e.context.get("error", e)`, so the old string doesn't even surface there. The formatter is the sole user-facing rendering path; changing the log string is out of the spec's message/formatter scope.

## Verification
- RED (`4ca45909`, test-only): 7 genuine assertion failures on the reframe; the discovery-sentinel guard stayed green (kills a "blanket add auto-discovered everywhere" cheat).
- GREEN (`fdfc0ae5`, source-only): full suite **7977 passed**, `ruff check`/`ruff format --check`/`mypy` clean.
- New/updated tests pin: file-in-title reframe, module fallback when no file path, cause preserved, `-u` + move-it-out next step in `fix`, and the sentinel title unchanged.
- Manual: rendered the spec's exact scenario through `format_user_error` — output matches the spec mockup; sentinel branch confirmed unchanged.
- Reward-hack note: the 6 old-title assertion swaps are a legitimate behavior change (title intentionally reframed) — net assertions increased (added auto-discovered/`-u`/move-out/cause checks + 2 new tests).
- Smoke skipped: formatter-only change, no chokepoint/registry surface (`risk.sh` smoke_required=no).
- Review: LOW risk → one blind fresh-context reviewer, VERDICT YES.

### DIFF
--- a/agent_actions/logging/errors/formatters/udf_load.py
+++ b/agent_actions/logging/errors/formatters/udf_load.py
@@ -108,9 +108,10 @@ class UDFLoadErrorFormatter(ErrorFormatter):
             )
 
         fix += (
-            "\n\nUDF discovery imports every Python file under the user-code "
-            "directory; a single broken file blocks all commands that load "
-            "the workflow."
+            "\n\nThis file was found by the recursive scan of the -u tree. Fix its "
+            "module-load code, or — if it is not a UDF — move it outside the -u "
+            "directory so discovery skips it. A single broken file blocks every "
+            "command that loads the workflow."
         )
 
         ctx: dict[str, Any] = {}
@@ -121,9 +122,17 @@ class UDFLoadErrorFormatter(ErrorFormatter):
         if requested_path:
             ctx["requested_path"] = requested_path
 
+        # Name the file, not the dotted module: discovery imported it from the
+        # -u tree, so "Failed to load UDF module 'X'" read as if the user chose
+        # it. Fall back to the module when no file path is available.
+        if file:
+            title = f"Auto-discovered UDF file failed to import: {file}"
+        else:
+            title = f"Auto-discovered UDF module failed to import: '{module}'"
+
         return UserError(
             category="Configuration Error",
-            title=f"Failed to load UDF module '{module}'",
+            title=title,
             details=details,
             fix=fix,
             context=ctx or None,
--- a/agent_actions/logging/errors/formatters/udf_load.py
+++ b/agent_actions/logging/errors/formatters/udf_load.py
@@ -108,10 +108,11 @@ class UDFLoadErrorFormatter(ErrorFormatter):
             )
 
         fix += (
-            "\n\nThis file was found by the recursive scan of the -u tree. Fix its "
-            "module-load code, or — if it is not a UDF — move it outside the -u "
-            "directory so discovery skips it. A single broken file blocks every "
-            "command that loads the workflow."
+            "\n\nThis file was found by the recursive scan of the user-code "
+            "directory (set via `-u <path>` or `tool_path` in your workflow "
+            "config). Fix its module-load code, or — if it is not a UDF — move "
+            "it outside that directory so discovery skips it. A single broken "
+            "file blocks every command that loads the workflow."
         )
 
         ctx: dict[str, Any] = {}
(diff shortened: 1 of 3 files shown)


Muizzkolapo:
## Real-project verification (qanalabs-quiz-maker)

Ran `agac list-udfs -u tools` against the real project, which already contains a genuinely broken auto-discovered UDF — `tools/run_thinkific_gen/apply_html_text.py` imports `markdown2`, which isn't installed. This is an organic reproduction of the exact bug, not a synthetic fixture (it's also the real failure the unit-test fixture was modeled on).

**Before** (formatter @ `d53d6817`, same real file):
```
Error: Configuration Error: Failed to load UDF module 'run_thinkific_gen.apply_html_text'
```

**After** (this branch, same real file):
```
Error: Configuration Error: Auto-discovered UDF file failed to import: tools/run_thinkific_gen/apply_html_text.py

  Problem: Python could not import the UDF module: No module named 'markdown2'
  File: tools/run_thinkific_gen/apply_html_text.py
  Fix: Python could not find the module 'markdown2'. If it is a third-party dependency, install it ...
  This file was found by the recursive scan of the -u tree. Fix its module-load code, or — if it is not a UDF — move it outside the -u directory so discovery skips it. ...
```

Title reframed to name the file; cause (`No module named 'markdown2'`) and the missing-module install guidance preserved; `-u`-tree next step present. The `list-udfs` command renders the shared formatter, so `inspect` / `validate-udfs` follow the same path. No changes left in the project (probe files removed, git-clean) or the formatter (restored to HEAD after the before/after swap).

Muizzkolapo:
## xhigh review triage (10 finders → 23 candidates → 11 reported)

Verified each finding against the code. Acted on the clear ones in `503681c7`:

**Fixed**
1. **`-u` hard-coded in the fix text** (strongest finding). Discovery runs from the `-u` path *or* `tool_path` config (`config_pipeline.py:106-111`); a `tool_path`-only user was told to "move it outside the -u directory" — a flag they never used. Restored source-agnostic wording naming both sources, matching the sibling `_format_discovery_failure` text in the same file.
2. **Noun mismatch in the file-absent title.** The `else` branch titled it a "module" while the shared fix suffix says "This file". Collapsed to one branch (`file or module`) — resolves the mismatch, drops a title variant no production raise site reaches, and removes the empty-string-truthiness edge.
3. **`Spec 570` in a test comment** — dropped the reference per CLAUDE.md §9 (kept the rationale).

**Deferred, with reason**
- **`str(exc)` in `configuration.py:114` still says "Failed to load UDF module 'X'"** (log surface) vs the new CLI title. The spec scoped `configuration.py` read-only, and the standard warning path uses `e.context.get("error", e)`, not the title — so the old string doesn't surface in normal logs. Left as-is to respect spec scope; happy to align it as a follow-up if log/CLI parity is wanted.
- **File path renders in both the title and the `File:` context line.** The spec explicitly requires the file in the title; `File:` is the structured-context anchor (with `search_path`/`requested_path`). Removing either regresses a spec requirement or a context test, so the mild redundancy is kept intentionally.

**Refuted / not acted on:** the `_SKIP_FIELDS` "module never rendered" concern (the *old* title already rendered the module; the skip-field only governs the context dict), the docstring/fix-text duplication (pure style; docstring `imports every Python file` is pre-existing 564-era rot, out of scope), and the empty-string truthiness (no producer emits an empty path; now moot after #2).

Full suite **7977 pass**, ruff clean after the changes.

PR #785, opened 2026-07-12
feat(udf): runtime data.require() strict accessor for RECORD-mode UDFs

Muizzkolapo:
## Summary
- Adds `Bus(dict[str, Any])` (`agent_actions/utils/udf_management/bus.py`): a transparent `dict` subclass. `get`/`[]`/iteration/membership stay tolerant; `require(namespace)` raises a `KeyError` naming the unknown namespace **and** the available ones — so RECORD-mode UDF authors can opt into fail-loud reads on a namespace typo. Runtime companion to spec 567's static bus-namespace scan.
- Wires it at the sole invocation point (`execute_user_defined_function`, `tooling.py`): RECORD-mode dict input is wrapped in a `Bus` before the UDF runs. FILE-mode (a `list`) is excluded via `isinstance(input_data, dict)`; already-wrapped inputs are a no-op. Imported lazily inside the function, matching the file's existing `registry` import convention.

## Deviations from the spec
- **Test layout split (harness-driven).** The spec's RED is an `ImportError` on the missing `Bus` in a single `test_bus.py`. The harness RED gate rejects collection errors (ImportError) as fake-RED and forbids the RED file changing between RED and GREEN. So the RED anchor is a genuine **assertion** failure at the wire point — a RECORD UDF calling `data.require("nope")` on a plain dict raises `AttributeError` whose wrapped message omits `"nope"` — in `test_bus_wiring.py` (no `Bus` import). The `Bus` unit tests live in `test_bus.py`, created at GREEN. Added `tests/unit/utils/udf_management/__init__.py` (package-style test tree).
- **Behavior change (documented, low risk):** `Bus(input_data)` is a shallow copy, so a UDF that rebinds a **top-level** bus key in place no longer leaks that mutation to the caller (nested mutations still propagate). Blast-radius review found zero real consumers that mutate UDF input in place; both callers (`llm/providers/tools/client.py:102`, `input/preprocessing/filtering/evaluator.py:202`) use only the return value/truthiness, and the guard path already treats its input as copy-safe.

## Review — HIGH tier (touches the `udf_management/` chokepoint), 3 diverse-lens blind reviewers
- **Correctness → YES**: `granularity` is the `Granularity` enum at the wrap; `Bus` preserves every relied-on dict behavior; FILE-mode excluded; double-wrap guard idempotent; grep found zero `type(x) is dict` strict-identity breaks.
- **Blast-radius → NO → fixed**: one LOW — `bus.py`/`Bus` missing from `_MANIFEST.md`; added. No bypass path (the only other `get_udf_metadata` users are static `inspect.getsource` analysis).
- **Anti-gaming → NO → fixed**: one MEDIUM — the FILE-mode exclusion guard was untested (a regression → `Bus([...])` `ValueError` crash); added `test_file_mode_input_is_not_wrapped`. Confirmed the suite is non-gameable: a trivial `def require(self, ns): return self[ns]` FAILS `test_require_error_lists_available_namespaces` (distinct names `ground`/`writer`/`typo`, no substring overlap), and the wiring assertion distinguishes "require ran and named the namespace" from "any exception."
- Both actionable issues fixed in `b784917f` (reviewed source logic unchanged); recorded `YES_WITH_ISSUES`.

## Verification
- **Harness gates:** RED = genuine assertion failure at the wire point; GREEN re-ran the RED target (hash-verified) + ruff + mypy + full suite **8022 passed**; risk.sh=HIGH → 3 blind reviewers; smoke **skipped** — `udf_management/` is not a smoke surface (surfaces: input/loaders, output/response, config, cli, llm, storage, prompt); the UDF-invocation path is exercised by the passing suite.
- **Real-project probe (`qanalabs-quiz-maker`):** the real RECORD-mode tool `aggregate_code_verification` (reads `data.get('compile_verify_input')`) ran unchanged through the wrap; the UDF now receives a `Bus` (`has_require=True`); `data.require('verify_code_answr_1')` — a real typo of `verify_code_answer_1` — failed loud through `execute`, naming the bad key and the available namespaces (`compile_verify_input`, `verify_code_answer_1`, ...). Project left git-clean (read-only import).
- 9 new tests (`test_bus.py` 6 unit incl. anti-gaming + mutation-copy; `test_bus_wiring.py` 3 invocation incl. FILE-mode no-wrap). `get`/`[]`/iteration unchanged.

Out of scope: the static preflight scan is spec 567; `data.get(...)` stays the tolerant path for genuinely optional reads.

### DIFF
--- a/agent_actions/utils/udf_management/_MANIFEST.md
+++ b/agent_actions/utils/udf_management/_MANIFEST.md
@@ -13,4 +13,6 @@
 | `clear_registry` | Function | Clears the registry (test cleanup) in a thread-safe manner. | `logging` |
 | `tooling.py` | Module | Utilities for loading/executing user-defined functions with schema validation support. | `errors`, `logging` |
 | `load_user_defined_function` | Function | Dynamically imports a module and returns the requested callable, searching sys.path if needed. | `errors` |
-| `execute_user_defined_function` | Function | Runs a UDF (optionally validates output against compiled schemas) and surfaces validation errors. | `errors` |
+| `execute_user_defined_function` | Function | Runs a UDF (RECORD-mode dict input wrapped in a `Bus`; optionally validates output against compiled schemas) and surfaces validation errors. | `errors` |
+| `bus.py` | Module | The action-name-keyed namespace dict handed to RECORD-mode UDFs, with a strict accessor. | — |
+| `Bus` | Class | `dict` subclass: `get`/`[]`/iteration stay tolerant; `require(namespace)` raises naming the unknown key and the available namespaces. | — |
--- a/agent_actions/utils/udf_management/bus.py
+++ b/agent_actions/utils/udf_management/bus.py
@@ -6,13 +6,14 @@ from typing import Any
 
 
 class Bus(dict[str, Any]):
-    """Action-name-keyed bus. `get`/`[]` stay tolerant; `require` raises on unknown."""
+    """Strict-accessor dict for UDF input. `get`/`[]` stay tolerant; `require` raises on unknown."""
 
     def require(self, namespace: str) -> Any:
+        # Message stays generic: the same accessor wraps the action-keyed tool
+        # bus and the flat guard-clause context, so it names the key + the
+        # available keys and does not assume an action-name keying.
         if namespace not in self:
             raise KeyError(
-                f"UDF read unknown bus namespace '{namespace}'. The bus is keyed by "
-                f"action name; available: {sorted(self.keys())}. Check for an "
-                f"impl-name/action-name mismatch."
+                f"UDF read unknown key '{namespace}'. Available keys: {sorted(self.keys())}."
             )
         return self[namespace]
--- a/agent_actions/utils/udf_management/bus.py
+++ b/agent_actions/utils/udf_management/bus.py
@@ -1,0 +1,18 @@
+"""Bus: the namespace dict passed to RECORD-mode UDFs, with a strict accessor."""
+
+from __future__ import annotations
+
+from typing import Any
+
+
+class Bus(dict[str, Any]):
+    """Action-name-keyed bus. `get`/`[]` stay tolerant; `require` raises on unknown."""
+
+    def require(self, namespace: str) -> Any:
+        if namespace not in self:
+            raise KeyError(
+                f"UDF read unknown bus namespace '{namespace}'. The bus is keyed by "
+                f"action name; available: {sorted(self.keys())}. Check for an "
+                f"impl-name/action-name mismatch."
+            )
+        return self[namespace]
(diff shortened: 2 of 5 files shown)


Muizzkolapo:
## Max-effort code review triage (47 agents, 10 finder angles → 34 candidates → 17 kept / 17 refuted)

Verified every kept finding against the code. The review's central, correct insight — missed by the pre-merge blind reviewers — is that the wrap fires at a chokepoint shared by **two semantically different callers**: the tool path (`client.py:102`, the action-keyed bus) and the guard `conditional_clause` path (`evaluator.py:202`, a **flat field-keyed** eval context from `_build_evaluation_context`). The root cause is that `Bus.require()`'s message assumed it only ever wraps the action bus.

### Fixed (commit `0daa567c`)
1. **[CONFIRMED] Misleading `require()` message on the guard path.** The old text ("keyed by action name; …impl-name/action-name mismatch") is factually wrong when `Bus` wraps the flat guard context. `Bus` is a generic strict-accessor dict — its message is now source-agnostic (`UDF read unknown key 'X'. Available keys: [...]`), accurate for any dict. The bus-specific guidance already lives in the static `bus_namespace_validator` (spec 567), so nothing is lost for the primary case. This also resolves the "duplicated guidance text" finding by removing the runtime copy.
2. **[CONFIRMED] Over-claiming comment.** `tooling.py` no longer asserts the input is always the action-keyed bus; it names both callers.
3. **[CONFIRMED] Test comment referenced the planning spec** (CLAUDE.md §9) — removed.

### Documented / accepted (not fixed, with rationale)
- **[CONFIRMED] `require()` KeyError swallowed by the guard `except Exception` (`evaluator.py:205`).** Not a regression — `require()` is a new capability; the guard path previously raised `AttributeError` (also swallowed). Guard clauses have their own passthrough-on-error semantics; `require()` fail-loud is a tool-path feature.
- **[CONFIRMED, cosmetic] `str(KeyError(msg))` repr-wraps the message** through `safe_format_error`. The spec's own adversarial review already weighed this and chose to keep `KeyError` (the tests mandate it); the namespace + available keys still surface.
- **[PLAUSIBLE] Shallow-copy nested-mutation leak / guard-path top-level mutation visibility.** Documented shallow-copy tradeoff (already in the PR body); blast-radius review found **zero** real UDFs that mutate input in place. Eliminating it would require dropping `require()` (a subclass needs a copy).
- **[PLAUSIBLE] None-vs-absent; str-context inputs; TrackedItem provenance.** `require()` checks key existence (the inverse of `get`'s tolerance) — a deliberate contract; non-dict inputs are out of the spec's dict scope; TrackedItem is FILE-mode/list-only (unreachable, the review agrees).

### Verification
Full suite **8023 pass**, mypy + ruff clean at `0daa567c`; real-project probe re-run on qanalabs `aggregate_code_verification` with the generic message (bad key + available keys surface). Smoke skipped (udf_management/ not a smoke surface).

PR #786, opened 2026-07-12
fix(validation): body-scan FILE-mode UDFs so filter mode is not flagged

Muizzkolapo:
## Summary

`agac validate-udfs` warned about every FILE-granularity UDF whose return annotation was not `FileUDFResult` — including legitimate **filter-mode** UDFs that return their input items unchanged as `list[dict]`, which the runtime accepts. That is noise: the runtime (`pipeline_file_mode.py`) only raises when a FILE-mode UDF returns a `list` containing freshly-built plain `dict` objects rather than the tracked input items.

This gates the warning on an AST **body-scan** that mirrors the runtime discriminator: warn only when the UDF **returns freshly-built dicts** — directly, as a list literal/comprehension of dict literals or `dict()` calls, or via a list accumulator (append/extend/assignment, including aliases) that those were fed into. Filter-mode UDFs, local lookup-table dict literals, and lists wrapped in `FileUDFResult(...)` are all silent.

### Why key on the return value, not on `append({...})` sites
The spec draft flagged any `X.append({...})`. But the idiomatic N→M UDF builds `outputs.append({"source_index": ..., "data": ...})` and then `return FileUDFResult(outputs=outputs)` — which is **safe**. Keying on append sites false-positives on it; keying on what is actually **returned** matches the runtime exactly (`FileUDFResult(...)` is a call, never a bare dict/list).

## Deviations from the spec (566 had already landed)
- The spec said "add tests," but spec 566 shipped with tests that asserted filter-mode UDFs (`return data`) **are** warned — the exact behavior 571 removes. Those 6 existing tests plus 2 in `test_validate_udfs.py` were **updated** to construct dicts where a warning is intended (they now assert corrected behavior — not weakened to pass). The green gate's reward-hack check flagged 5 removed assert/def lines; this is that legitimate update.
- The draft `_constructs_new_dicts` is replaced by a return-based scan that also closes its documented false negatives: `dict()` calls, build-in-a-var-then-append, and dict comprehensions are now detected.
- A blind review found one further false negative — a list accumulator aliased before return (`out = results; return out`) — now closed via fixpoint alias propagation, pinned by a flagged-alias test and a filter-alias guard test.

## Known limitation
Static heuristic: merge/expand via un-modelled shapes (e.g. `collections.OrderedDict(...)`, a dict returned from a helper call) is not flagged. The runtime check remains the backstop; UDFs whose source can't be read are left unflagged.

## Verification
- `pytest tests/validation/test_file_udf_contract_validator.py tests/validation/test_validate_udfs.py` — pass.
- Full suite: **8024 passed**; `ruff check` / `ruff format --check` clean; `mypy agent_actions` clean.
- **Real project** (`qanalabs-quiz-maker`): ran the detection logic against all 9 real FILE-mode UDFs — **0 false positives**, including `convert_html_json_to_thinkific` (unannotated, appends dict literals to `outputs`, returns `FileUDFResult(outputs=outputs)`) which the draft would have wrongly flagged.
- Smoke ring skipped: validation-only static AST check, no chokepoint/runtime/project-I/O surface touched.

### DIFF
--- a/agent_actions/validation/file_udf_contract_validator.py
+++ b/agent_actions/validation/file_udf_contract_validator.py
@@ -1,7 +1,11 @@
-"""Flag FILE-granularity UDFs whose return annotation is not FileUDFResult."""
+"""Flag FILE-granularity UDFs that return freshly-built dicts without FileUDFResult."""
 
 from __future__ import annotations
 
+import ast
+import inspect
+import textwrap
+from collections.abc import Callable, Iterator
 from typing import Any
 
 from agent_actions.config.types import Granularity
@@ -9,10 +13,8 @@ from agent_actions.utils.udf_management.registry import FileUDFResult
 
 
 def _returns_fileudfresult(annotation: Any) -> bool:
-    # Coarse check by design: class identity, or the annotation's text mentions
-    # FileUDFResult (covers string forward-refs and Optional/Union wrappers). It
-    # reads only what is already recorded — never imports or resolves the
-    # annotation, which could fail.
+    # Text-match by design: never imports or resolves the annotation (that can
+    # raise); covers the class, string forward-refs, and Optional/Union wrappers.
     if annotation is FileUDFResult:
         return True
     if isinstance(annotation, str):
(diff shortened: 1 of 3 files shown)


Muizzkolapo:
## Real-project verification (live `agac validate-udfs`)

Ran the actual command against `qanalabs-quiz-maker` (both workflows reference FILE-mode UDFs: `convert_html_json_to_thinkific`, `deduplicate_across_documents`, `assign_batch_name`), full discovery + registry + config-expansion path.

**Before (main / 566)** — `agac validate-udfs -a qanalabs_quiz_gen -u tools`:
```
✅ All UDF references valid
⚠ FILE-mode UDF 'convert_html_json_to_thinkific' returns <class 'inspect._empty'>, not FileUDFResult. ...
```
This is a false positive: the UDF actually does `return FileUDFResult(outputs=outputs)` (safe) — it just lacks the return annotation.

**After (this branch / 571)** — same command, both `qanalabs_quiz_gen` and `qanalabs_quiz_gen_batch`:
```
✅ All UDF references valid
  - 23 Tools referenced in config
  - 67 Tools discovered and registered
```
No FILE-mode warning. The body-scan reads the in-body `return FileUDFResult(...)` as safe. Exit 0, no tracebacks. The warning is gone because the UDF is genuinely safe — not because the check went inert (unit tests prove a real constructor still warns).

Note: an unrelated pre-existing blocker (`tools/run_thinkific_gen/apply_html_text.py` imports the uninstalled `markdown`/`markdown2`) reproduces identically on `main`; installing those two genuine project deps unblocked discovery for this comparison.

PR #789, opened 2026-07-13
fix(prompt): name the seed_path namespace in the missing-seed-data error (578)

Muizzkolapo:
## Summary

The framework has two similarly-named seed-data knobs that collide in the error path:

- The **folder** that holds seed files is configurable via `seed_data_path` in `agent_actions.yml` (default `seed_data`).
- The **context namespace** actions read seed data through is hardcoded to `seed_path`.

Because `seed_data` (folder default) and `seed_path` (namespace) look alike, a user conflated them and set `seed_data_path: seed_path`, expecting that to line the two up. Their files were still in `seed_data/`, so every run died at the first action with an opaque `Seed data directory not found. Create 'seed_path' folder …` — no indication that `seed_path` is a separate namespace concept, and no hint that a `seed_data/` folder already existed. That cost a full debugging session.

### Change (Task 1 decision: Option B)

Kept the `seed_path` namespace unchanged and fixed the guidance:

- The missing-seed-data error now names the resolved folder, states the folder name comes from `seed_data_path`, and clarifies that actions always reference seed data under the `seed_path` context namespace — a separate knob.
- When `seed_data_path` is set to `seed_path` **and** a conventional `seed_data/` folder exists (the exact reported footgun), the error appends a targeted hint to remove the misconfiguration.
- Extracted `_find_workflow_root` so the folder walk-up is shared between resolution and the footgun check (no duplicated search loop).

**Why B, not A:** Option A (make the namespace track `seed_data_path`) is a behavior change for every project already using the `seed_path` namespace and would need a compat window supporting both namespace names at once — a dual-code-path/backward-compat shim, which is merge-blocking here. B satisfies the full success metric with a message + targeted-hint change and no change to the happy path.

### Deviation from the spec

- The spec floated an *optional* "warn at load" for the `seed_data_path == seed_path` footgun. Implemented as an **error-time** hint on the single on-path raise site instead of a separate config-load warning — it lands exactly where the user hits the wall and avoids expanding into the config-validation layer. Same detection condition the spec named (`seed_data_path == "seed_path"` while a `seed_data/` folder exists).

## Verification

- **RED → GREEN:** new `tests/unit/prompt/test_seed_data_missing_message.py`
  - message names the folder + `seed_path` namespace + `seed_data_path` knob;
  - footgun hint fires only when `seed_data_path=seed_path` and `seed_data/` exists;
  - negative test asserts the hint is absent otherwise;
  - happy-path test asserts the default `seed_data/` folder resolves with no error.
- Full suite: **8039 passed**; `ruff format --check`, `ruff check`, `mypy` clean.
- The three pre-existing `_determine_static_data_dir` regression tests still pass after the `_find_workflow_root` extraction (walk-up semantics preserved: fallback to the config file's parent when no `agent_config/` is found).
- Blast radius: the only `Seed data directory not found` site in production is the one changed; `_determine_static_data_dir` has a single caller (`_load_seed_data`); no sibling instances.
- Blind review (LOW tier): **VERDICT YES**.
- **Smoke:** `scan-project.sh` registry/side-effect canary **passes** (registries match the manifest). The online smoke (`agac run` on qanalabs) was **skipped**: the run reached a live external LLM HTTPS call and hung 8+ min with zero output — it executed all code this diff touches (config load, prompt prep, seed resolution) and stalled on the external provider, i.e. environmental/host-specific, not a branch fault. `baseline.sh` has no timeout of its own, so it would reproduce the identical unbounded hang and risk leaving the repo on `main` — not an informative confirmation. The diff's only behavioral surface (missing-seed error message + a regression-test-verified refactor) is fully covered by the unit suite above.

### DIFF
--- a/agent_actions/prompt/service.py
+++ b/agent_actions/prompt/service.py
@@ -717,22 +717,24 @@ class PromptPreparationService:
             logger.warning("Error during seed data resolution: %s", e, exc_info=True)
             # Fall through to error raising
 
-        # Not found - raise an error that names the folder AND the seed_path namespace.
+        # Not found - name the folder and distinguish it from the seed_path
+        # context_scope directive (the common point of confusion).
         message = (
-            f"Seed data directory not found. Create '{seed_dir_name}' folder "
-            "at workflow root (same level as agent_config/, schema/, prompt_store/) "
-            "to store static reference data files. The folder name comes from "
-            f"'seed_data_path' in agent_actions.yml (currently '{seed_dir_name}'); "
-            "actions always reference seed data under the 'seed_path' context "
-            "namespace, regardless of the folder name."
+            f"Seed data directory not found. Looked for the '{seed_dir_name}' folder "
+            "at the workflow root (same level as agent_config/, schema/, prompt_store/). "
+            "The folder name defaults to 'seed_data' and can be changed with "
+            "'seed_data_path' in agent_actions.yml; it is separate from the 'seed_path' "
+            "directive in an action's context_scope, which only lists the seed files to "
+            "load. Create the folder to store static reference data files."
         )
         if seed_dir_name == "seed_path" and PromptPreparationService._default_seed_folder_exists(
             workflow_config_path
         ):
             message += (
-                " A 'seed_data' folder already exists — 'seed_path' is the context "
-                "namespace, not the folder name. Remove 'seed_data_path: seed_path' "
-                "from agent_actions.yml to use the default 'seed_data' folder."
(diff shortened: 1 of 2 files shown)


Muizzkolapo:
Superseded by spec 583 (seed-directive rename). #789 was the error-message band-aid for the seed_path vs seed_data_path collision; 583 removes the collision at the root by renaming the context_scope directive seed_path -> seed (and retiring the dead static_data key), so the seed_path-specific footgun hint here becomes obsolete. The still-useful part — the 'seed data directory not found' message naming the folder and the seed_data_path knob — is folded into 583 in corrected form.

PR #791, opened 2026-07-14
fix(context): rename the seed_path directive to seed; retire static_data

Muizzkolapo:
## Summary

"Static reference data available to prompts" was fragmented across **five** similar names, and the directive key collided with the folder knob — a user set `seed_data_path: seed_path` (conflating the two) and lost a debugging session (the origin of #578).

| Name | Role |
|------|------|
| `seed` | the namespace prompts read — `{{ seed.x }}` |
| `seed_path` | context_scope directive that declares which files to load *(renamed here → `seed`)* |
| `static_data` | a second, half-wired directive for the same thing *(removed here)* |
| `seed_data` | default folder name (+ a flagged misuse token) |
| `seed_data_path` | folder-location config knob *(unchanged)* |

The framework already carried two band-aids for this confusion (an anti-typo linter, and #578's error message). This collapses the declaration to **one** name that matches the reference:

- **Rename** the directive `seed_path` → `seed` (declare `context_scope.seed:`, read `{{ seed.* }}`).
- **Remove** the dead `static_data` directive (was in config types + shown by `agac inspect`, never read by the runtime loader).
- **Fail loud** on the removed keys: `normalize_context_scope` raises a `ConfigurationError` pointing at `seed` instead of silently loading no seed data. This runs on the runtime config-load path (`config/manager.py`) and the validation path — so a leftover `seed_path:` errors clearly rather than silently dropping seeds.
- The old names stay in `SEED_CONFIG_KEYS` so writing `seed_path.x` in a prompt reference is still flagged as misuse.

The folder knob `seed_data_path` and the `seed` namespace are unchanged. **Clean cutover** — no dual-key acceptance, no deprecation shim (per the repo's no-compat-shim rule; one real project migrated).

## Deviations from the spec

- **Chose fail-loud** for the removed keys. The spec left "error vs silently load nothing" as an implementation decision; the no-silent-fallback rule makes erroring (with migration guidance) the correct choice, so the change adds a `REMOVED_DIRECTIVES` guard rather than relying on the prior silent-preserve behavior.
- **Superseded #578 / PR #789** (the error-message band-aid), which is now closed — the `seed_path`-vs-folder collision it patched no longer exists.
- **Left internal `static_data` plumbing** (the loaded-data parameter/variable in `scope_application`/`service`, the `cache_type="static_data"` event label, the `seed_path` `Path` local in `lsp/resolver`). These are internal names, not the directive; renaming them is out of scope and would expand blast radius (event schema, etc.).

## Verification

- **RED → GREEN:** `tests/unit/prompt/test_seed_directive.py` (loads under the new `seed` key; `seed_path`/`static_data` fail loud). Full suite **8070 pass**; `ruff format --check`, `ruff check`, `mypy` clean.
- **Fixture migration is a pure rename** — the gate's reward-hack flag ("22 removed test lines") is a false positive: every removed `assert`/`def` line was a `seed_path`→`seed` rename (confirmed by diff audit and by the anti-gaming reviewer); no assertion was weakened.
- **Review (risk=HIGH → 3 diverse-lens blind reviewers):**
  - Correctness/mechanism → **YES**.
  - Blast-radius → **NO → fixed**: caught stale docs still showing the old directive (`prompt`/`output`/`skills`/`llm` `ARCHITECTURE.md`, `agac-agent-skills/SKILL.md`, `static_loader`/`realtime builder` docstrings/comments) — all updated so copying them can't trigger the new error.
  - Tests/anti-gaming → **YES**; its notes addressed: extracted `ContextCommand._seed_namespace_keys` + a test that `agac inspect` surfaces the `seed` namespace, and strengthened the removed-key tests to assert the error names both the old key and the `seed` replacement.
  - Recorded **YES_WITH_ISSUES** after remediation.
- **Smoke (skipped, offline real-project proof instead):** `scan-project.sh` canary **passed**. On the **migrated qanalabs config**, `agac inspect -a qanalabs_quiz_gen` passes preflight for every action (config normalization + seed resolution succeed under `seed:`; fail-loud guard correctly not tripped), and `agac inspect context` surfaces the `seed` namespace with all 5 keys rendering as `{{ seed.* }}`. The online `agac run` is environmentally blocked on the external LLM (hangs) and adds nothing for the paths this diff touches.
- **Migration:** the qanalabs workflow configs were migrated `seed_path:` → `seed:` (prompts already read `{{ seed.* }}`).

## Pre-existing (not introduced here)
- Verbose-prose flags at `service.py:479`, `static_loader.py:2`, `builder.py:47`, and three test-file module docstrings are pre-existing in files this rename touched.

### DIFF
--- a/agent_actions/cli/inspect_action.py
+++ b/agent_actions/cli/inspect_action.py
@@ -464,11 +464,10 @@ class ContextCommand(BaseInspectCommand):
         context_scope = action_config.get("context_scope") or {}
 
         # `seed.*` is populated by the `seed` directive.
-        seed_entries = context_scope.get("seed")
-        seed_keys: list[str] = list(seed_entries.keys()) if isinstance(seed_entries, dict) else []
+        seed_keys = self._seed_namespace_keys(context_scope)
         # Don't shadow a user action literally named `seed`.
         if seed_keys and "seed" not in namespaces:
-            namespaces["seed"] = list(dict.fromkeys(seed_keys))
+            namespaces["seed"] = seed_keys
 
         return {
             "action_name": self.target_action_name,
@@ -481,6 +480,14 @@ class ContextCommand(BaseInspectCommand):
             },
             "total_template_variables": sum(len(fs) for fs in namespaces.values()),
         }
+
+    @staticmethod
+    def _seed_namespace_keys(context_scope: dict[str, Any]) -> list[str]:
+        """Keys exposed under the ``seed.*`` namespace, from the ``seed`` directive."""
+        seed_entries = context_scope.get("seed")
+        if not isinstance(seed_entries, dict):
+            return []
+        return list(dict.fromkeys(seed_entries.keys()))
 
     @staticmethod
     def _group_refs_by_namespace(refs: list[Any]) -> dict[str, list[str]]:
--- a/agent_actions/cli/inspect_action.py
+++ b/agent_actions/cli/inspect_action.py
@@ -463,13 +463,9 @@ class ContextCommand(BaseInspectCommand):
 
         context_scope = action_config.get("context_scope") or {}
 
-        # `seed.*` is populated by the `seed_path` and `static_data` directives.
-        # (`seed_data` in SEED_CONFIG_KEYS is an anti-typo flag, not a real key.)
-        seed_keys: list[str] = []
-        for src_key in ("seed_path", "static_data"):
-            entries = context_scope.get(src_key)
-            if isinstance(entries, dict):
-                seed_keys.extend(entries.keys())
+        # `seed.*` is populated by the `seed` directive.
+        seed_entries = context_scope.get("seed")
+        seed_keys: list[str] = list(seed_entries.keys()) if isinstance(seed_entries, dict) else []
         # Don't shadow a user action literally named `seed`.
         if seed_keys and "seed" not in namespaces:
             namespaces["seed"] = list(dict.fromkeys(seed_keys))
--- a/agent_actions/config/types.py
+++ b/agent_actions/config/types.py
@@ -43,8 +43,7 @@ class ContextScopeDict(TypedDict, total=False):
     passthrough: list[str]
     drop: list[str]
     keep: list[str]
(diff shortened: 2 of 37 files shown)


Muizzkolapo:
## xhigh code-review remediation (commit 73b1bb8a)

A workflow-backed xhigh review found **two real regressions** the harness's blind reviewers missed, plus doc/test stragglers. All fixed:

**Regressions:**
1. **6 bundled example workflows broke at load** — my rename sweep covered `agent_actions/`, `tests/`, and the qanalabs project but not `examples/`. All six example configs (`book_catalog_enrichment`, `contract_reviewer`, `incident_triage`, `product_listing_enrichment`, `review_analyzer`, `support_resolution`) still declared `context_scope.seed_path:`, so the new fail-loud guard aborted every one. Migrated the configs + their READMEs `seed_path:` → `seed:`.
2. **The static analyzer crashed on a legacy directive** — `WorkflowStaticAnalyzer.analyze()` called `normalize_context_scope` unguarded, so `agac validate`/`inspect` on an un-migrated config hit a raw `ConfigurationError` instead of a collected, formatted error. Now caught and recorded as a `StaticTypeError` so the pass still reports the rename alongside the workflow's other issues. (Fail-loud stays for the *runtime* path.)

**Stragglers/tests:** doc references still naming the retired directive (`llm/ARCHITECTURE.md` diagram, `scope_builder` field-context comment, `validation/ARCHITECTURE.md` misuse list, a bogus `{static.field}` in the realtime builder comment); a test that the misuse check flags `static_data.field`; and a test that `analyze()` reports (not raises on) a legacy directive.

**Deferred with rationale:** generalizing "all unknown directives raise" (a larger behavior change with its own blast radius); `seed:` null → `{}` normalization (pre-existing dict-directive behavior, no live failure). **Refuted (agreed):** DRY/style nits.

Verification: **8072 tests pass**, ruff + mypy clean. Note: `agac inspect` on the `contract_reviewer` example still fails on a *pre-existing, unrelated* missing `templates/` dir — not caused by (or in scope for) this rename.

PR #788, opened 2026-07-12
feat(source_guid): unify first-stage generation on a deterministic content identity

Muizzkolapo:
> **Stacked on #787** (base = `fix/source-guid-invariant`). #787 enforces `source_guid` *presence*; this enforces its *determinism*. Review the delta only; merge after #787.

## Summary
`source_guid` is the framework's cross-run record key — the disposition/checkpoint gate matches on it (`disposition_gate.py:70,122`). But first-stage generation minted a **random `uuid4`**, so:
1. **Batch checkpoint/incremental resume was silently broken** — a re-run minted new keys, so dispositions never matched.
2. **The dual-stamp "must-agree" contract was broken** — a record is keyed twice (once for `source_data` persistence, once for processing/disposition; the code documents this at `initial_pipeline.py:514`), and random generation made them disagree, so the `source_data` row and its disposition didn't join.

**Fix:** one shared `IDGenerator.derive_source_guid(record)` — deterministic `uuid5` over the record's **content**, excluding volatile framework envelope fields (`target_id`, `batch_uuid`, `batch_id`, `node_id`, `*_target_id`, `source_guid`). Every first-stage stamp site routes through it (`initial_pipeline` batch+online, `task_preparer`, `online_llm`, `unified`'s first-stage backfill), replacing the random calls. Same input content → same guid across runs and across stamp sites, **by construction**. Random `generate_source_guid()` stays only for genuinely-new entities (expansion children `enrichment.py:70`, FILE-mode synthetics) — out of scope.

## Verification
- **Harness:** genuine RED (batch first-stage guid non-deterministic today); GREEN full suite **8042 pass** + ruff + mypy; risk.sh=HIGH → 3 diverse-lens blind reviewers.
- **Real-project capstone** (qanalabs, all 129 staging records): two runs of the same input → **identical `source_guid`s**, distinct per record, **ingestion guid == processing derive**, and `target_id` still varies per run (proving determinism is content-driven, not a frozen envelope).
- **Reviewers:** anti-gaming **YES**, blast-radius **YES**, correctness **NO → fixed**.

## Reviewer findings — all addressed
[...]

PR #792, opened 2026-07-14
fix(source_guid): hash raw content at ingestion, drop name-based exclusion (#577)

Muizzkolapo:
## Summary

`derive_source_guid` computed a first-stage record's identity by **subtracting framework field names** (`node_id`/`target_id`/`batch_id`/…) from the record. A user CSV/JSON column that happens to share one of those names — e.g. a graph/edge export keyed on `node_id` — was stripped from the hash, so two genuinely-distinct rows derived the **same** `source_guid` and one was silently dropped by `INSERT OR IGNORE` on `UNIQUE(relative_path, source_guid)`. (Follow-up F2 from #788.)

**Root cause:** identity was derived by name-based subtraction because the batch stamp sites hashed the record *after* the volatile envelope was merged in, then guessed which keys were framework by name — a guess that fails when user data shares a name.

**Fix — identity is hashed over the raw payload, before the envelope, with no subtraction:**
- `generator.derive_source_guid` hashes the payload as given; nothing is stripped by name.
- `initial_pipeline` batch stamp sites (`_prepare_text_chunks_batch`, `_add_batch_metadata`) derive over the raw content/row **before** the envelope is added, so the per-run `target_id`/`batch_id` never enter the guid.
- `record/envelope.SOURCE_GUID_EXCLUDED_FIELDS` **deleted** — obsoletes 576's single-authority-for-the-exclusion-set apparatus (there is nothing left to exclude).
- `task_preparer` site 6 (the one re-derive that would hash an *enveloped* record) **fails loud** on a missing first-stage guid instead of fabricating a poisoned one (extends #787's no-fallback posture).
- `online_llm.py` first-stage exception handler now **reads** the stamped guid instead of re-deriving (review-caught regression — see below).

## Why it's safe (strict no-op for real data)

Hashing the raw payload reproduces the exact old value wherever a row has no reserved-named column — verified against all four frozen 576 golden guids (`865e85c1…`, `cac9d8de…`, `6aaf98d9…`, `bd78edf7…`). The `source_guid` changes **only** in the collision case. The ~30 read-side consumers (disposition gate, checkpoint resume, second-stage source_resolution, version_correlation_id) read a value that does not move.

## Verification

- **RED → GREEN, harness pipeline.** RED genuine (two collision tests failed on real assertions); GREEN full suite **8059 pass**, ruff + mypy clean.
- **End-to-end on the real project (qanalabs-quiz-maker, ql_code_debug).** Injected two distinct records differing only in `node_id` and ran the real pipeline:
  - Before the fix: **2 records in → 1 stored** (one silently dropped), exit 0, no warning.
  - After the fix: **2 → 2**, distinct guids, user's `node_id` preserved in storage, exit 0.
- **risk.sh = HIGH** (chokepoint `record/`), smoke_required=no → 3 diverse-lens blind reviewers + 1 confirming re-verify.

## Review (HIGH tier)

3 blind reviewers: blast-radius **YES**, tests/anti-gaming **YES**, correctness **NO → fixed**. The correctness reviewer caught a regression this change introduced: the online first-stage generic-exception handler re-derived `source_guid` over an item that `unified.py:114` had already stamped in place — with stripping gone, the stamped guid entered the hash and diverged from the persisted key (`e3ebebf1…` vs `798ac46b…`), so a FAILED disposition would be keyed on a guid absent from `source_data` and re-run matching would break. Fixed by **reading** the stamped guid (matching the sibling failure paths `_build_prep_failed_result`, `preparator.py`), not by re-adding an exclusion (which would have masked the caller). Added a non-tautological regression test. A fresh blind verifier re-checked the new HEAD.

## Test changes (reward-hack note)

The diff removes ~27 lines from existing tests and deletes `test_source_guid_single_authority.py`. This is legitimate, not weakening-to-pass: the deleted/rewritten assertions pinned the *old subtraction contract* (e.g. `test_excludes_volatile_envelope_fields` asserted two records differing in `node_id` get the SAME guid — the bug), and the deleted file tested `SOURCE_GUID_EXCLUDED_FIELDS`, which no longer exists. The "no persisted guid moved" property is preserved and strengthened (golden pins moved to `test_derive_source_guid.py` covering all four historical values), plus new reserved-collision, cross-run-determinism, and online-exception-path regression tests.

## Deviations / follow-ups (not in this PR)

- **Batch envelope still overwrites a user column named like a framework field in the *stored* record** (value loss, distinct from the identity collision this PR fixes). The deeper cure is namespacing the record envelope — a large cross-cutting change with no current workload hitting it; deferred to its own spec.
- **Pre-existing (not a regression here):** JSON-scalar / XML-non-list batch inputs and online tabular/XML loaders don't stamp `source_guid` at all (tracked separately, spec 579).


Muizzkolapo:
### Confirming blind re-verify (correctness) at HEAD `39f5e428` — VERDICT: YES

A fresh-context blind correctness reviewer re-checked the branch after the review-response fix. Summary of what it verified by execution:

- **No-op for real data:** reproduced main's old name-strip derivation vs the new raw-payload derivation side-by-side — `OLD == NEW == golden` for every golden payload; no persisted `source_guid` moves.
- **Bug + fix:** two rows differing only in `node_id` — OLD collapses to one guid, NEW yields distinct guids, through the real `_add_batch_metadata`.
- **Ingestion ↔ processing agreement** confirmed by execution for batch, online-JSON, and online-text.
- **`online_llm` fix:** confirmed re-deriving over an already-stamped record now diverges, and the fix (read the stamped guid) yields the persisted `source_data` key.
- **Fail-loud raise:** reachable only on a malformed first-stage-shaped record wrongly routed as non-first-stage; `DataValidationError` falls through to the per-record generic handler → FAILED, never an action-fatal crash. Cannot fire on a legitimate path.

**One informational note (reviewer judged correctness-safe, not a bug):** for a first-stage online-*text* record (raw string) that hits a *generic* exception in `task_preparer.prepare`, the handler now yields `source_guid=None` (main derived a guid), so no FAILED disposition is written and the record is reprocessed on re-run rather than carried as failed. This writes **no** key, so the "persisted guid == disposition key" invariant is not violated, and it matches the two sibling failure handlers (`_build_prep_failed_result`), which already return `None` for raw strings.

Muizzkolapo:
Superseded by #793 (spec 584 — the record-envelope namespace boundary). 584 subsumes this PR's identity fix (still hashes the raw payload before the envelope) and adds the structural fix that makes BOTH the collision AND the value overwrite impossible by construction: the first-stage user payload lives under content.source, framework fields stay flat. Closing in favor of #793.

PR #793, opened 2026-07-14
fix(envelope): namespace first-stage user payload under content.source (#584, supersedes #792)

Muizzkolapo:
## Summary

**The root-design fix for the reserved-field bug class — supersedes #792.** First-stage BATCH records used to spread the user row FLAT alongside framework fields (`{**row, batch_id, target_id, node_id, source_guid, ...}`), so framework metadata and user data shared one namespace. Two consequences:

1. A user column named like a framework field (`node_id`/`target_id`/`batch_id`/…, e.g. a graph/edge export keyed on `node_id`) was **overwritten** in storage (value loss).
2. The whole flat record — framework ids included — **leaked into the `source.*` namespace**.

\#792 fixed only the *identity collision* (hash the raw payload before the envelope). 584 does the structural fix that **subsumes** it: the user payload lives under its own `content.source` namespace; framework fields stay flat. Now framework and user data **cannot collide by construction** — value preserved, no leak, and identity is still hashed over the raw payload (which *is* `content.source`), so no `source_guid` moves. This PR contains #792's commits as its identity foundation; **#792 is closed as superseded.**

**Option B, no backwards compat** (dev, disposable stores). First-stage guards now use `source.<field>` — the evaluator already promotes `content`, so this Just Works; bare field names no longer resolve, which is the intended clean break *and* converges batch to the already-documented online contract (removes a batch↔online divergence).

## Changes (the spike's exact migration list — 4 sites)

- `initial_pipeline._add_batch_metadata` / `_prepare_text_chunks_batch`: emit `{content: {source: {**row}}, ...framework flat...}`; derive `source_guid` over the raw payload **before** wrapping (guid unchanged).
- `initial_pipeline._should_save_source_items`: compare the user payload keys (`content.source`) via a new `_source_payload_keys()` helper, not the fixed framework top-level keys.
- `scope_application._resolve_source_content`: return `content.source` so FILE-mode `source.<field>` resolves to the user field, not `{"source": {...}}` (the double-nesting the spike caught). Scoped to that one caller — `_extract_content_data`'s other caller (`scope_builder.py:133`) is unaffected (`source` is skipped there via `SPECIAL_NAMESPACES`).

Most of the pipeline **already speaks `content.source`** (it's the downstream record convention), so only 2 existing tests pinned the old flat shape (updated to `content.source`).

## Why it's safe

- **Value preserved & no leak** — proven on the real 129-record qana batch input: `source.*` resolves to exactly the 6 user fields, zero framework leak; a user `node_id` survives.
- **No persisted `source_guid` moves** — identity is hashed over the raw payload; the frozen golden `cac9d8de-…` reproduces.
- **Collision-proof by construction** (reviewer-verified): a user column named `source` is rejected at ingestion (`SPECIAL_NAMESPACES`); `content.source` can't collide with `content[action_name]` (`source` is a reserved agent name); a user column named `content` nests safely at `content.source.content`.

## Verification

- **RED → GREEN**, harness pipeline. RED genuine (value-preservation + framework-leak, both fail on the flat base). GREEN full suite **8068 pass**, ruff + mypy clean.
- **Real project (qanalabs_quiz_gen_batch, 129-record input):** batch ingestion + `source.*` resolution verified on real data. Full LLM batch run blocked by a **pre-existing** qana config error (`StaticTypeError: action 'generate_source_caveats' does not exist`) + env-blocked ollama — `baseline.sh` confirms the workflow fails identically on `main`; my diff touches zero config/preflight code.
- **risk.sh = HIGH → 3 diverse-lens blind reviewers, all VERDICT YES:** correctness (live-traced all goals + full collision analysis), blast-radius (every consumer — delta storage, batch result strategy, `scope_builder.py:133`, guard eval — SAFE; guard bare-field = parity fix), tests/anti-gaming (every wrong-impl caught). Their test-coverage gaps (literal-shape assertion, task-prep guid inheritance, `content`-named column) are now closed by `test_first_stage_envelope_shape.py`; a FILE-mode guard test proves `source.*` resolves against the wrapped envelope (fails without the `_resolve_source_content` fix).
- **Smoke skipped:** `scan-project` canary diverges on **pre-existing shared-manifest drift** (added qana UDFs) — `baseline.sh` confirms it fails identically on `main`; not `--refresh`ing a manifest other clones depend on. Shape change verified via the real-data probe + guard tests instead.

## Test changes (reward-hack note)

The diff removes ~35 lines from existing tests. Legitimate: the deleted assertions pinned the OLD flat shape (`result[0]["name"]` → `result[0]["content"]["source"]`), strengthened to exact-payload equality; the identity-only #792 test rewrites are included. No assertion weakened to pass.

## Behavioral change (intended, no back-compat)

First-stage BATCH guards using a **bare** user-field name now require `source.<field>`. This matches the documented namespaced contract and the existing online behavior (online already required the dotted path), removing a batch↔online divergence. All in-repo guard tests use dotted paths.

## Deferred (pre-existing, out of scope — all 3 reviewers confirmed)

- The single-object JSON (`_prepare_json_batch` non-list) and non-list XML batch branches emit an unwrapped, `source_guid`-less record — a correct fix needs dict-vs-scalar handling + `source_guid` stamping (579-adjacent). Identical on `main`, untouched here.
- Online tabular/XML loader `source_guid` stamping (spec 579).


Muizzkolapo:
### Review: 3 diverse-lens blind reviewers + confirming re-verify — all VERDICT: YES

- **Correctness** — live-traced all four goals (value preserved, no framework leak, source_guid unchanged, source.* resolves in RECORD+FILE mode) and a full collision analysis: a user column named `source` is rejected at ingestion (`SPECIAL_NAMESPACES`); `content.source` can't collide with `content[action_name]` (`source` is a reserved agent name); a user column named `content` nests safely at `content.source.content`.
- **Blast-radius** — every consumer SAFE (delta storage `_extract_delta`, `batch_result_strategy`, `scope_builder.py:133` via `SPECIAL_NAMESPACES` skip, guard evaluator, checkpoint/disposition writers). The first-stage-guard bare-field change is a batch→online **parity correction**, not a regression.
- **Tests/anti-gaming** — every wrong-impl caught (still-overwrites, still-leaks, moved-guid, FILE-mode double-nesting, RECORD-mode). Flagged 3 coverage gaps → now closed by `test_first_stage_envelope_shape.py` (literal `content.source` shape; `content`-named column; task-prep guid inheritance, non-tautological) + the FILE-mode double-nesting guard.
- **Confirming re-verify (final HEAD)** — YES; investigated and dismissed a false-positive `reprompt_ops.py` `user_content` flag (a framework LLM-message field, byte-identical at base `39f5e428`, unaffected by 584).

Deferred (pre-existing, all reviewers confirmed out-of-scope, byte-identical at base): single-object-JSON / non-list-XML batch branches emit an unwrapped `source_guid`-less record (579-adjacent).

Muizzkolapo:
### Real batch `agac run` on the qanalabs project — the wrap verified end-to-end on disk

The only existing batch workflow (`qanalabs_quiz_gen_batch`) is blocked at preflight by a pre-existing config error (dangling `generate_source_caveats` / `assemble_quiz_options` actions), so I ran a minimal batch workflow in the qanalabs project (branch code, `run_mode: batch`, two records differing only in a reserved `node_id` column), then removed the scaffold.

`source_data` written by the real batch ingestion:

```
top keys: [batch_id, batch_uuid, content, node_id, parent_target_id, root_target_id, source_guid, target_id]
content.source = {id: r1, page_content: "alpha body", node_id: "USER_A"}   guid dd32d531
content.source = {id: r2, page_content: "alpha body", node_id: "USER_B"}   guid f4cd4470
```

Confirmed through a genuine `agac run` in batch mode:
- **Wrap applied** — user payload under `content.source`, framework fields flat.
- **User `node_id` preserved** — `USER_A`/`USER_B` under `content.source`, not overwritten by the framework `node_id` (which is flat at top level).
- **No framework leak** — `content.source` holds only the 6 user fields.
- **No collision** — two rows differing only in `node_id` → distinct `source_guid`s, both stored.
- **Batch task prep resolved `source.*`** into the submitted batch job (the LLM action observing `source.page_content`/`source.node_id` built and submitted the batch task — it would have crashed at prep if resolution were broken).

(An earlier tool-only variant ran in online mode — a tool-only workflow has no LLM call to batch — and independently confirmed `source.node_id` resolves to the user's value end-to-end.)

Muizzkolapo:
### Extended to ONLINE mode — both paths now use one pathway

Per maintainer direction ("online and batch must use the same pathway"), the boundary now covers online ingestion too. Online JSON and text branches (`_prepare_online_data`) now emit the **identical** `{content: {source: {…}}, source_guid}` shape as batch; the processor **inherits** the stamped guid (identity still hashed over the raw item, so no guid moves), and `get_existing_content`'s name-based-strip synthesis branch is now **dead** for online first-stage.

**Why it was needed:** online used to emit FLAT records, so `get_existing_content` stripped a reserved-named user column (`node_id`) by name and `source_guid` leaked into `source.*` — the same disease, unfixed on the online path.

**Blast radius:** essentially the two `_prepare_online_data` branches. Because the batch work already made every downstream consumer `content.source`-aware, this needed **no downstream changes** and broke **zero** existing tests (8077 pass).

**Real-project proof (online `ql_code_debug`, `node_id` column):** `source_data` now wrapped, `node_id`=`USER_A`/`USER_B` preserved, no `source_guid` leak, distinct guids — was flat with `node_id` stripped before.

**Review (HIGH tier, 3 diverse-lens on the online delta):** correctness **YES** (traced JSON+text through the processor live; source_guid unchanged incl. falsy-guid edge; all re-derive sites inherit; text str→dict safe), blast-radius **YES**, tests **NO→fixed** (both reviewers found the same gap: tests asserted on `src_text` but the processor consumes `data_chunk` — a wrap-only-src_text regression would pass; closed by pinning `data_chunk is src_text` + processor-inherit tests). Recorded **YES_WITH_ISSUES**.

**Out of scope (pre-existing, all reviewers confirmed):** tabular/XML online branches (`.csv/.tsv/.xlsx/.xml`) — the 579 loader-stamping gap, tracked separately.

PR #804, opened 2026-07-17
fix(557): clear stale node-level disposition on execute, drop read-side heal

Muizzkolapo:
## Summary

- New helper `_clear_stale_node_disposition` runs at the top of `_execute_action_run` (and the async variant), wiping any prior-round `NODE_LEVEL_RECORD_ID` row for this action.
- Deletes the read-side heal at `executor.py:718-727` — node-level disposition is now guaranteed to reflect the current round only, so the heal is unreachable.
- Simplifies `_resolve_completion_status` to return `SKIPPED` unconditionally when a `NODE_LEVEL_RECORD_ID` `SKIPPED` row is present — that row can only come from the current round's `_handle_dependency_skip` or `_handle_all_versions_filtered`.
- Per-record dispositions are untouched (`clear_disposition` is keyed on `record_id`).

Eliminates the operator-visible warning that fired on every rerun after a prior dependency-skip:
```
Stale guard-skip disposition on 'auto_review_quality' — action has 1 target file(s).
A write path set SKIPPED despite output existing. Clearing disposition and proceeding as COMPLETED.
```

Spec: `specs/active/557-stale-node-disposition-clear-on-execute.md`.

## Root cause

`SQLiteBackend.set_disposition` only DELETEs prior rows for its own `(action_name, record_id)` before INSERTing. When an action in round N wrote only per-record dispositions (no NODE_LEVEL write), a prior-round `SKIPPED@NODE_LEVEL` row from `_handle_dependency_skip` survived, and `_resolve_completion_status` reconciled the "SKIPPED but has output" contradiction with a scary warning. The right layer to enforce "NODE_LEVEL rows reflect the current round only" is the moment the executor commits to running this round.

## Test removals (reward-hack disclosure)

Three obsolete tests were removed because they pinned the deleted heal's behavior. Per the spec: "If a test asserts the heal block still runs, do not restore the block." Removed:

- `test_stale_skipped_disposition_cleared_when_output_exists` — asserted the heal cleared the SKIPPED disposition when target files existed. Behavior deleted.
- `test_action_produces_output_but_has_stale_skipped_resolves_completed` — same behavior, integration framing.
- `test_database_error_from_list_target_files_propagates` — `_resolve_completion_status` no longer calls `list_target_files`, so this specific propagation site is gone. Sibling storage-error tests (`test_operational_error_from_has_disposition_propagates`, `test_operational_error_from_get_failed_items_propagates`) still cover the propagation invariant for the two calls that remain.

`test_returns_skipped_when_all_records_guard_skipped` was renamed to `test_returns_skipped_when_node_level_skipped_disposition_present` and updated to reflect the new invariant (no `list_target_files` check).

## Verification

- 5 new regression tests in `tests/workflow/test_stale_node_disposition.py`:
  - prior-run SKIPPED cleared when current run executes (sync)
  - no `Stale guard-skip disposition` warning fires on rerun
  - per-record dispositions survive the node-level clear
  - `_handle_dependency_skip` in the current round is NOT cleared (guardrails against wrong insertion point)
  - async `_execute_action_run_async` clears too
- Full suite: **8139 passed** (0 failed).
- `ruff format --check` clean, `ruff check` clean, `mypy` clean.
- Blind code reviewer verdict: **YES** — traced every writer of NODE_LEVEL disposition, confirmed clear-on-execute fires strictly before every in-round writer, and every "prior-run" writer that could reach `_resolve_completion_status` is now blocked.
- Smoke skipped: `risk.sh` says `smoke_required=no`; diff touches only `workflow/executor.py`, not in `smoke-surfaces.txt`.

## Deviations from spec

- Spec line numbers were stale (referenced `executor.py:1051-1074` and `586-595`); the file has grown to 1279 lines. Same functions and heal block, correct sites patched: helper at `1187`, sync call at `1215`, async call at `1251`, heal removed from `703-729`.
- Spec's expected `_resolve_completion_status` shape kept the `if not target_files: return SKIPPED` branch. Deleted the `list_target_files` check entirely — with clear-on-execute enforcing the invariant, a surviving SKIPPED row means this round's dependency-skip and always maps to SKIPPED. The simpler shape falls out cleanly.
- Also updated one existing test (`test_returns_skipped_when_all_records_guard_skipped` in `test_circuit_breaker.py`) whose implementation-detail assertion (`list_target_files.return_value = []`) no longer fits the code path.

## Out of scope

- Extending `_cleanup_stale_dispositions` to handle SKIPPED across runs — separate larger discussion; not needed once clear-on-execute lands.

### DIFF
--- a/.changes/unreleased/Bug Fix-20260717-557000.yaml
+++ b/.changes/unreleased/Bug Fix-20260717-557000.yaml
@@ -1,0 +1,3 @@
+kind: Bug Fix
+body: "Clear node-level disposition when an action begins executing. Eliminates the 'Stale guard-skip disposition' warning that fired on reruns after a prior dependency-skip, and removes the read-side heal block in _resolve_completion_status."
+time: 2026-07-17T20:00:00.000000Z
--- a/agent_actions/workflow/executor.py
+++ b/agent_actions/workflow/executor.py
@@ -708,23 +708,13 @@ class ActionExecutor:
         if storage_backend.has_disposition(
             action_name, DISPOSITION_SKIPPED, record_id=NODE_LEVEL_RECORD_ID
         ):
-            target_files = storage_backend.list_target_files(action_name)
-            if not target_files:
-                logger.info(
-                    "Action '%s' had all records guard-filtered — marking as skipped",
-                    action_name,
-                )
-                return ActionStatus.SKIPPED
-            storage_backend.clear_disposition(
-                action_name, DISPOSITION_SKIPPED, record_id=NODE_LEVEL_RECORD_ID
-            )
-            logger.warning(
-                "Stale guard-skip disposition on '%s' — action has %d target file(s). "
-                "A write path set SKIPPED despite output existing. "
-                "Clearing disposition and proceeding as COMPLETED.",
+            # Clear-on-execute guarantees this row is current-round: only
+            # _handle_dependency_skip could have written it, so map directly to SKIPPED.
+            logger.info(
+                "Action '%s' had all records guard-filtered — marking as skipped",
                 action_name,
-                len(target_files),
             )
+            return ActionStatus.SKIPPED
         item_failures = storage_backend.get_failed_items(action_name)
         if item_failures:
             if not storage_backend.has_successful_items(action_name):
@@ -1191,8 +1181,30 @@ class ActionExecutor:
             metrics=ExecutionMetrics(duration=duration),
         )
 
+    def _clear_stale_node_disposition(self, action_name: str) -> None:
+        """Enforce the invariant that NODE_LEVEL disposition reflects the current round only.
+
+        Per-record dispositions (real source_guid keys) are unaffected — the
+        clear is keyed on record_id.
+        """
+        storage_backend = getattr(self.deps.action_runner, "storage_backend", None)
+        if storage_backend is None:
+            return
+        try:
+            storage_backend.clear_disposition(
+                action_name=action_name,
+                record_id=NODE_LEVEL_RECORD_ID,
+            )
+        except Exception as err:
+            logger.warning(
+                "Failed to clear stale node-level disposition for %s: %s",
+                action_name,
+                err,
+            )
+
     def _execute_action_run(self, params: ActionRunParams) -> ActionExecutionResult:
         """Execute action run (synchronous)."""
+        self._clear_stale_node_disposition(params.action_name)
         self.deps.state_manager.update_status(params.action_name, ActionStatus.RUNNING)
         self._track_action_start(params)
         try:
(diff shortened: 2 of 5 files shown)


Muizzkolapo:
Live end-to-end verification on the real qanalabs DB (`agent_workflow/code_centered_quiz/agent_io/store/code_centered_quiz.db`), added after initial submit:

- Installed this branch as editable framework in the qanalabs venv (`uv pip install -e`).
- Seeded the exact bug scenario on the two actions with target data (`code_extractor`, `prescreen_code_blocks`): `SKIPPED@__node__` disposition + `target_files` present.
- Drove `_execute_action_run` directly against each (no LLM cost).
- Both stale rows cleared; 0 `Stale guard-skip disposition` warnings emitted (previously would fire once per action per rerun).
- DB restored from backup — no residual state.

**Smoke-surfaces false negative to note:** `workflow/executor.py` isn't in `harness/config/smoke-surfaces.txt`, so `risk.sh` set `smoke_required=no`, but the change is functionally observable in the real store. Consider adding `agent_actions/workflow/` to the smoke-surfaces list in a follow-up.

Muizzkolapo:
Follow-up commit `dae71e84` addresses the three self-review findings:

1. **(HIGH) Silent-swallow in `_clear_stale_node_disposition`** — removed. With the read-side heal deleted, a silently-caught `clear_disposition` failure re-created the pre-fix bug (stale row survives, resolver now trusts it, run misclassified as SKIPPED). Storage errors on clear now propagate loud, same posture as spec 554's `_count_records_for_action`. New regression test `test_storage_error_during_clear_propagates` pins it. Docstring updated to state the invariant + reference the sibling precedent.

2. **(Cleanup, LOW)** — the try/getattr/log-warning parallel with `_write_failed_disposition`/`_write_skipped_disposition` — intentionally not extracted. The terminal writers keep best-effort semantics (writing at a terminal state, run already ending); the clear now has fail-loud semantics (writing at run start, invariant load-bearing). Consolidating the shape would force one policy on both — a false unification.

3. **(Test-quality, LOW) `test_returns_skipped_when_node_level_skipped_disposition_present`** — added `list_target_files.assert_not_called()`. Guardrail against accidentally regrowing the deleted heal's cross-check.

Suite: **8140 pass** (was 8139 + 1 new test), ruff+mypy clean.

Muizzkolapo:
Second-round review follow-up (commit `1ba00b80`):

1. **Resolver comment** at `executor.py:711` no longer mis-attributes the surviving `SKIPPED@NODE_LEVEL` row. Both `_handle_dependency_skip` and `_handle_all_versions_filtered` return before `_resolve_completion_status` runs — the real writer is `result_collector.write_node_level_disposition` firing during `run_action` when every input record was filtered. Comment now names the right writer and calls out the two that don't reach the resolver so a future maintainer can find them.

2. **CLAUDE.md convention** — removed `spec 554` citations from the executor docstring and the new test docstring; the reasoning stays, the ticket reference doesn't (rule: don't reference tasks/PRs in code).

3. **Clear scope docstring** — `_clear_stale_node_disposition` now says explicitly that it wipes every NODE_LEVEL disposition type for the action, not just SKIPPED; the 'current-round only' invariant is what makes that safe.

Suite: **8140 pass**, ruff+mypy clean.

PR #808, opened 2026-07-18
fix(593): enforce the schema-echo reject invariant at the checkpoint + online-result seams

Muizzkolapo:
## Summary

Closes the two remaining seams where a record whose action namespace is the framework's
own compiled output schema (`is_schema_echo`: `type:object` + `properties` + `title`)
could persist as a `success`. 556 already gates the `target_data` seam (`write_target`)
and `_reject_schema_echo_items` gates the executed-LLM response; these were the two
uncovered paths:

1. **Checkpoint seam** — `save_checkpoint_records` (`sqlite_backend.py`) was ungated.
   It now runs the same content-nested gate `write_target` uses: a schema-echo namespace
   is replaced with the parse-error sentinel and recorded FAILED/PARSE_ERROR before it is
   checkpointed, so a resume / carry-forward read never picks it up as valid output. The
   shared base-class gate is seam-agnostic, so it is renamed
   `_gate_schema_echo_deltas` -> `_gate_schema_echo_records` (private, one caller).

2. **Online result funnel** — `OnlineLLMStrategy.invoke` appended each `process_record`
   result straight to collection/checkpoint. It now passes every result through
   `_reject_schema_echo_result` first: a schema-echo namespace from **any** branch becomes
   the parse-error sentinel and the result becomes **FAILED**, keeping the in-memory status
   and the persisted data/disposition consistent (never a silent success). This moves the
   catch to the result seam so status and persistence agree, rather than relying on the
   persistence gate to silently rewrite a still-`success` result's data.

The invariant is now total across `target_data`, `checkpoint_output`, and the online
result path. (The batch path already gates independently at `batch_result_strategy.py:315`.)

## Deviations from the spec (593)

- **Record shape.** The spec's RED sketches show top-level `{"<action>": compiled}` records.
  The real checkpoint/result shape is **content-nested** `{"content": {"<action>": compiled}}`
  (produced by `transform_with_passthrough` -> `RecordEnvelope.build`; confirmed by
  `disposition_gate` reading `read_checkpoint_records` and `read_target` interchangeably).
  Both gates therefore read `record["content"][action_name]`, and the tests use the
  content-nested shape. Gating the spec's stated top-level key would have been a silent
  no-op. (Captured in `tasks/lessons.md`.)
- **No live planter found (expected).** Per the spec's own framing this is a deterministic
  invariant closure, not a live-repro fix. No current no-LLM branch emits schema-echo data;
  RED-2 constructs the offending result to prove the seam. The known 125 corrupt rows
  predate the 556 gate and are historical + contained.

## Verification

- **RED->GREEN, two cycles** (harness): each RED recorded a genuine assertion failure
  before its fix; GREEN re-ran the RED target + full suite + ruff + mypy.
  - Checkpoint seam: `tests/unit/storage/test_schema_echo_checkpoint_gate.py` (real SQLite
    backend, write + read-back + disposition).
  - Online seam: `tests/unit/processing/test_online_schema_echo_seam.py` drives through
    `invoke` -> real `_checkpoint_record` -> real `save_checkpoint_records` read-back. The
    result-status assertion was hardened (per review) to pin exactly FAILED + a positive
    `failed` disposition + the persisted `_parse_error` -- proven non-tautological by
    re-recording RED against a disabled gate.
- **Full suite:** `pytest` 8172 passed; `ruff format --check` + `ruff check` + `mypy agent_actions` clean.
- **556 regression:** the 7 `test_schema_echo_gate.py` tests and the executed-LLM guard
  tests stay green; the title-present shape is still detected everywhere.
- **Fixtures are real:** built via `compile_unified_schema(..., "ollama_cloud")`, i.e. the
  actual title-present compiled shape, not a guessed literal.
- **Review:** HIGH tier, 3 blind diverse-lens reviewers (correctness, blast radius, tests)
  all confirmed the fix correct (3/3). The tests-lens raised one non-tautological-assertion
  gap, fixed in the final code commit.
- **Real-project smoke: justified skip (baseline-confirmed pre-existing).** The smoke ring
  is environmentally blocked on `main` too: (1) the scan-project canary is stale (the
  qanalabs project gained `m_*/mc_*/ql_*/to_qanalabs_*` UDFs since the manifest was
  recorded -- this diff does not touch UDF discovery); (2) `smoke-test.sh online` fails at
  config validation with `context_scope.seed_path is not a valid directive` before any
  storage/online code runs. `baseline.sh` verdict for both: pre-existing/environmental.


Muizzkolapo:
Addressed the blocking review note: stripped the per-clone \`tasks/\` scratch from the PR.

- Added commit untracks \`tasks/lessons.md\` (which was erroneously tracked on \`main\` — the pull-abort/re-promotion cause flagged in coordination/status.md) and drops the branch-new \`tasks/todo.md\`; both working copies are retained on disk so \`promote-lessons.sh\` still picks up this run's lesson into \`shared/lessons/\` on \`/harness done\`.
- \`.gitignore\` now ignores \`tasks/\` to prevent recurrence.

Net PR diff is now the fix (3 source + 2 tests), the \`.gitignore\` guard, and the \`tasks/lessons.md\` untrack. The "(Captured in tasks/lessons.md)" note in the description refers to that local, now-untracked scratch — it will land in \`shared/lessons/lessons.md\` via promotion, not this PR.

PR #820, opened 2026-07-28
fix: fan-in grouping keys on one universal field, not per-record first-present (UAT DP audit)

Muizzkolapo:
## Summary
- **Fixes silent partial fan-in merges** (UAT DP audit, DP-06 — pre-marked known bug): `merge_records_by_key` selected the correlation key independently per record (first present of `version_correlation_id` / `root_target_id` / `parent_target_id` / `source_guid`). Branches of the same logical record carrying different key sets could never group — a version-merge-descendant branch keyed on its correlation id while a plain sibling branch keyed on its source guid — so the fan-in consumer received duplicate half-records, each missing the other branch's namespace, with no error. A second instance of the same class: expansion siblings sharing `parent_target_id` (no vc id) all keyed on the shared parent and silently collapsed into one record.
- **Fix:** the grouping key is chosen once for the whole pool — the first candidate present on every keyed record — with identity keys (`version_correlation_id`, `source_guid`) ranked before lineage keys (`root_target_id`, `parent_target_id`). Explicit `reduce_key` keeps its per-record precedence. Per-record selection remains the fallback when no candidate is universal, so disjoint-key pools behave as before.
- **Root-cause deviation from the audit doc:** DP-06 attributed the failure to *unequal branch depths*. Verified against the real project store: depth is a red herring — the actual trigger is key-space asymmetry between branches. An unequal-depth diamond with symmetric keys (e.g. `select_best_code_snippet`: draft depth-1 + three versioned generators depth-2) merged correctly in the real June-23 run because every branch carried an aligned inherited `version_correlation_id`.

## Verification
- RED (`tests/unit/workflow/test_merge_fanin_key_asymmetry.py`): asymmetric branches split into 4 half-records (expected 2); expansion siblings collapsed 3→1. Both genuine failures recorded by the gate before the fix.
- GREEN: full suite **8250 passed**, ruff clean, mypy clean. All 61 existing merge tests unchanged and green (including `reduce_key` precedence, parent-only grouping, keyless/non-dict passthrough).
- **Real-data replay** (qana_quiz June-23 store, `consolidate_quality_review`'s 4-branch fan-in, 20 records):
  - Unmodified pool: 5 merged records × 4 namespaces — byte-identical grouping before/after the fix (no regression on real shapes).
  - Asymmetric pool (one branch stripped of its vc id, simulating a branch that never crossed a version merge): **main → 10 records (5×3-namespace + 5×1-namespace halves); fixed → 5 records × 4 namespaces.**
- Blind fresh-context review (LOW tier, diff + bug statement only): see review verdict in the harness record.
- Smoke: skipped per `risk.sh` (`smoke_required=no` — no registered project surface touched; the change is exercised by the suite's runner/merge/data-bus integration tests).

## DP audit classification (context for this PR)
- DP-01 dependency wait, DP-02 skip propagation, DP-03 filter cascade (incl. sibling-branch re-entry guard), DP-04 multi-dep fan-in delivery, DP-05 cycle detection: all verified on current main — evidence in `coordination/status.md`.
- DP-06: fixed here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/agent_actions/workflow/merge.py
+++ b/agent_actions/workflow/merge.py
@@ -180,32 +180,68 @@ def _merge_group_deep(group: list[dict[str, Any]]) -> dict[str, Any]:
     return merged
 
 
+_AUTO_KEY_CANDIDATES = [
+    "version_correlation_id",
+    "source_guid",
+    "root_target_id",
+    "parent_target_id",
+]
+
+
+def _has_correlation_key(record: dict[str, Any], key_name: str) -> bool:
+    if record.get(key_name):
+        return True
+    content = record.get("content")
+    return isinstance(content, dict) and bool(content.get(key_name))
+
+
+def _select_universal_key(records: list[dict[str, Any]]) -> str | None:
+    """First auto candidate present on every record that carries any candidate.
+
+    Grouping must key every record in the pool on the SAME field: branches of
+    one logical record can carry different key sets (a version-merge descendant
+    has ``version_correlation_id``, a plain branch does not), and per-record
+    key selection would place them in disjoint groups — a silent partial merge.
+    Returns None when no candidate is universal (caller falls back to
+    per-record selection).
+    """
+    keyed = [r for r in records if any(_has_correlation_key(r, k) for k in _AUTO_KEY_CANDIDATES)]
+    if not keyed:
+        return None
+    for key_name in _AUTO_KEY_CANDIDATES:
+        if all(_has_correlation_key(r, key_name) for r in keyed):
+            return key_name
+    return None
+
+
 def merge_records_by_key(records: list[Any], reduce_key: str | None = None) -> list[Any]:
     """Merge records sharing the same correlation key.
 
-    For same-source fan-in (no ``reduce_key``), uses ``merge_branch_records``
-    so each branch contributes only its own namespace and upstream is preserved
-    from the base record (first record in the group).  For ``reduce_key``
-    aggregation (different sources grouped together), uses ``deep_merge_record``.
-
-    Note: When using the branch-records path, ``group[0]`` is the canonical
-    upstream source.  Its shared namespaces survive; other records' versions
-    of the same upstream namespaces are ignored.
+    The grouping key is chosen once for the whole pool (first universal
+    candidate, identity keys before lineage keys) so branches carrying
+    different key sets still group by their shared identity; an explicit
+    ``reduce_key`` takes precedence per record.  Same-source fan-in merges
+    via ``merge_branch_records`` with ``group[0]`` as the canonical
+    upstream (its shared namespaces survive); ``reduce_key`` aggregation
+    deep-merges groups instead.
     """
     groups_by_key: dict[str, list[dict[str, Any]]] = {}
     records_without_key: list[Any] = []
 
-    key_candidates: list[str] = []
-    if reduce_key:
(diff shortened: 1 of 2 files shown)


Muizzkolapo:
## Live E2E check (post-review, user-requested)

Ran a fresh `agac run -a ql_code_centered --fresh` against the real project with this branch on `PYTHONPATH` (July-26 baseline store backed up first). Outcome and what it proves:

- **Exit 0, but the run stalls at action 7/42 (`filter_by_syllabus`) — proven project-side, not a regression of this PR.** The workflow config has `dedup_code_blocks` observing only `describe_code_canonical.*` while four downstream actions observe `dedup_code_blocks.code_block/description/language`. The July-26 baseline only ever advanced ~13/55 records past that gate, and those 13 rode on the pre-#819 double-nesting leak (their `describe_code_canonical` namespace embeds a nested `flatten_code` dict, confirmed in the baseline store). On the clean current framework — main or this branch — there is no leak, so every record hits zero-overlap and the DAG skip-cascades. `merge_records_by_key` is never called before the stall (single-dependency chain), so this PR's change is not involved.
- **Framework behaved loudly at every layer:** startup preflight named the exact hazard ("prompt references `dedup_code_blocks.code_block` but producer does not guarantee it"), runtime emitted zero-overlap null-safe warnings per record, prep failures were dispositioned, and the skip cascade stated its reasons.
- **What did run on branch code ran clean:** 6 actions including the FILE-granularity embedding dedup (49→42 with correct `FileUDFResult` reconciliation) and the identical 6 deterministic `flatten_code` empty-output failures as the baseline.
- **Fan-in validation evidence for this PR remains:** the real-store replay in the PR body (unmodified 4-branch pool byte-identical before/after; asymmetric pool main→10 half-records vs fixed→5 fully-merged) plus the suite's runner/merge integration tests. The live path to `merge_code_alternatives`/`aggregate_code_verification` is blocked by the project config gap above until the one-line observe fix lands project-side.

Project follow-up filed in coordination status: extend `ql_code_centered.yml:59` dedup observe with specific `flatten_code.{code_block,description,language,complexity,applicable_context}` refs (specific refs, not a second wildcard — two wildcard namespaces deliver namespace-qualified keys by contract).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Muizzkolapo:
## Live end-to-end proof (follow-up to the check above)

Built a minimal all-tool project (no LLM calls) with the exact asymmetric fan-in: a versioned branch (`vote_1/2` → `merge_votes`, whose outputs carry `version_correlation_id`) and a plain branch (`plain_note`, no vc), both feeding `fanin_consumer`. Ran it through the full real pipeline (`agac run`, sqlite store, delta reconstruction, multi-dir fan-in merge) twice with identical inputs:

- **main (`96cd491e`)**: `fanin_consumer` processed **6 records for 3 items** — every item duplicated; the plain-branch half-record has no `merge_votes` namespace (`both_branches: false`). The silent duplicate/partial fan-in, live.
- **this branch (`b5d96d9a`)**: exactly **3 records**, one per item, all `both_branches: true`, same guids and correlation ids.

Repro project: 5 deterministic tool actions, 3 staging records, runs in ~0.5s — happy to check it into `examples/` or `tests/` as a fixture if wanted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #845, opened 2026-08-07
refactor: remove the unused dependency-injection layer

Muizzkolapo:
## Summary
Stacked on #844. Removes the unused dependency-injection layer (~700 LOC): `config/di/` (container, configurator, application, types) and `config/factory.py`, verified ceremonial end to end:

- The container **bindings were never resolved** and the `DIConfig` profiles (dev/prod/test) were never read
- `ProcessorFactory` was threaded `ActionRunner → strategies → ProcessingPipeline`, **null-checked at three layers, and never invoked**
- The `ProcessorRegistry` was **write-only**: three decorators registered components (`data_processor`, `data_generator`, `source_data` loader) that nothing ever looked up — registrations removed with it
- `service_init` now constructs `ActionRunner` directly; the ceremonial parameter is gone from every signature (`ActionRunner.__init__`, `ActionStrategy`, `PipelineConfig`/`ProcessParams`, both pipeline factories; the zero-caller `create_processing_pipeline` is deleted)
- `config/interfaces.py` is **untouched** — its ABCs are live contracts (the survey's zero-ref claim was wrong for `ILoader`/`IProcessor`, which are live bases)

**Bonus root-cause fix:** the `workflow` package barrel now resolves its exports lazily (PEP 562, same pattern as `input/loaders`), so importing a leaf submodule no longer executes the coordinator import chain. This kills the long-standing circular import (`…services.processing → processing.unified → workflow/__init__ → coordinator → … → processing.unified`) that made several test modules fail to collect standalone — `pytest tests/unit/test_batch_resubmission_guard.py` now collects and passes on its own.

## Verification
- Full suite: **8,178 passed** — delta vs parent is exactly the 42 tests in the three deleted DI test files plus the consolidated strategy repr/eq tests, all named in the diff
- ruff + mypy clean; surface diff vs parent = the five deleted modules, the two removed write-only `registry` symbols, and `~` on the six classes that lost the ceremonial parameter — every line accounted for
- Standalone collection of previously-uncollectable test modules verified working


Muizzkolapo:
Blind fresh-context review (diff + change statement only): **YES_WITH_ISSUES** — reviewer independently reproduced the circular import at the pre-fix commit and verified it resolved at HEAD; verified all six behavior-preservation claims (bindings never resolved, ProcessorFactory never invoked, registry write-only, direct construction equivalent, PEP 562 barrel safe — no eager side effects in coordinator/models/schema_service, DependencyError removal clean); zero missed siblings repo-wide. The two flagged stale doc lines (root manifest + config architecture still saying 'dependency injection') are fixed in the follow-up commit on this branch.

PR #854, opened 2026-08-08
fix: interrupted runs record terminal state instead of stalling at running

Muizzkolapo:
## Summary

A run killed by Ctrl-C or SIGTERM left its in-flight action reading `running` in `.agent_status.json` forever, and the workflow reading `running` in `.manifest.json`. Anything consuming those files — `agac status`, the VS Code sidebar — reported a live run long after the process was gone. `reset_retryable()` only clears it on the *next* run, so the lie persisted exactly while the user was looking at it.

**Root cause.** Both coordinator execution loops caught `except Exception`. The CLI signal handler (`cli/main.py:98`) raises `SystemExit`, and cancellation arrives as `CancelledError`; neither is an `Exception` subclass, so both walked past `handle_workflow_error` and the terminal-status sweep never ran. The remediation function already existed — the fix is reaching it.

**Why a new status.** Routing interrupts through `FAILED` destroys resume progress. `_reset_retryable_actions` keys its preserve-checkpoints branch on `RUNNING`, treating it as "died mid-processing, may hold checkpointed SUCCESS dispositions"; `FAILED` is bulk-wiped by design. Collapsing the two would make a Ctrl-C at record 9,000/10,000 reprocess all 10,000 — full LLM re-spend. `workflow/ARCHITECTURE.md:414` documents this invariant. `ActionStatus.INTERRUPTED` plus `MID_PROCESSING_STATUSES = {RUNNING, INTERRUPTED}` keeps both meanings: `RUNNING` still covers ungraceful death where nothing rewrote the status.

Ordering also matters. State is persisted **before** the failure event is fired: event dispatch does real disk I/O and `EventManager.fire` only contains `Exception`, so a second Ctrl-C landing inside a handler previously escaped before anything was written — reproducing the original bug with no diagnostics. And a manifest write failure is now logged rather than allowed to replace the interrupt (an ENOSPC turned Ctrl-C into exit 1 instead of 130).

## Verification

- **RED → GREEN**, re-recorded at each correction. 13 interrupt tests: `KeyboardInterrupt`, `SystemExit`, async-loop cancellation, manifest sweep, second-interrupt-during-event-dispatch, `BATCH_SUBMITTED` survives, ordinary-`Exception` regression guard, and cross-run checkpoint survival.
- The resume tests drive the **real** `_reset_retryable_actions`. The existing `test_coordinator_checkpoint.py:18-37` reimplements that method in a local helper, which is why it stayed green while the behaviour it guards was inverted.
- **Full suite: 8146 passed.** `ruff check`, `ruff format --check`, `mypy agent_actions` clean.
- `npm run typecheck` clean for `vscode-extension/src`. It caught two real defects in this diff: `Record<ActionStatus, ThemeColor>` was non-exhaustive, and `StatusSummary` had no counter for the new status (which would have gone `NaN`). The ~170 remaining errors are pre-existing, all in `agent_actions/tooling/docs/frontend/`, untouched here.
- **Smoke skipped**, justified: `risk.sh` reports `smoke_required=no` — the diff touches no path in `harness/config/smoke-surfaces.txt`.

Review was HIGH tier (coordinator chokepoint), three blind lenses over two rounds. They found the checkpoint regression, the event-ordering gap, a tautological test of mine, and five stale doc surfaces; all are fixed here.

## Deferred, deliberately

- **Per-action `CancelledError` inside `asyncio.gather(return_exceptions=True)`** (`parallel/action_executor.py:245`) is captured and only logged, so it never reaches the new handler. Top-level Ctrl-C during an async run *is* covered — `asyncio.run`'s shutdown throws `CancelledError` into the awaited coroutine. Nothing in the repo cancels individual action tasks today, so this is an unguarded path rather than a live failure. Fixing it means reworking the gather semantics, which does not belong in this diff.
- **Concurrent `agac run` on one workflow** can have one process's interrupt sweep another's live actions — `.agent_status.json` has no cross-process lock. Pre-existing on the error path; this change makes it reachable by Ctrl-C. VIOL-0045's advisory lock addresses it but is not on `main`.
- **`has_any_failed()`** counts only `FAILED`, so a fully-interrupted run would classify as SUCCESS — unreachable today, since interrupts re-raise past that check and resume resets to `PENDING`. Left alone rather than adding unreachable handling.


Muizzkolapo:
Note for merge: please **squash-merge**. The net diff is correct, but commit `8d730d43` is an intermediate state that routes interrupts through `FAILED` and would destroy checkpointed dispositions on resume — a bisect through this branch would hit it. Squashing keeps that out of `main`'s history.

PR #859, opened 2026-08-08
feat: tell a live run from a corpse by asking the OS

Muizzkolapo:
## Summary

Closes the last hole in the "sidebar claims a run is live when it isn't" report.

An earlier layer made Ctrl-C and SIGTERM record a terminal status via the CLI signal handler. But a process killed outright — SIGKILL, OOM, power loss — runs no handler at all, so `.agent_status.json` keeps claiming `running` with nothing to correct it. The spinner never stops. And since the workflow rollup sorts running first, a dead run pins itself above genuinely failed ones.

Neither file can answer this alone, because both are written by the process that died. So record **who that was**: `initialize_manifest` stamps `pid` and `hostname`, and the extension asks the OS whether that process still exists.

`process.kill(pid, 0)` performs the existence and permission checks without delivering a signal. `EPERM` counts as alive — a process owned by another user is still a process.

## Deliberately conservative

Liveness is `unknown`, and **nothing is relabelled**, when:

- the manifest predates this field (older `agac` against a newer extension),
- it was written on another host (`hostname` mismatch — a pid from another machine is meaningless),
- there is no manifest at all.

Only a definite `dead` downgrades `running`/`checking_batch` to `interrupted`. The failure mode that would be *worse* than the bug — a live run mislabelled as interrupted — requires the OS to report the owning process as absent while it is running, which `EPERM` handling covers.

`batch_submitted` is deliberately left alone: that work continues in the provider's queue after `agac` exits, so a dead local process says nothing about it.

## Verification

- RED recorded on 4 manifest tests before the field existed; then GREEN.
- **8170 Python tests**, ruff + mypy clean.
- **56 extension tests**, 12 of them covering `runLiveness`/`reconcileWithLiveness` — including every `unknown` path, that terminal statuses are untouched, and that `batch_submitted` survives.
- `npm run typecheck` clean for `vscode-extension/src`; build succeeds.

## Relationship to VIOL-0045

This does not replace the advisory `flock` on branch `fix/viol-0045-concurrent-run-lock`, which prevents concurrent runs. An OS lock is the stronger liveness signal — the kernel releases it even on SIGKILL — and would subsume this pid check if that branch lands. It is not on `main`, so this does not depend on it.

## Known limits

- **Pid reuse.** A recycled pid belonging to an unrelated process reads as live. That is the pre-change behaviour (spinner keeps spinning), so it is a false-negative, never a false-positive.
- **Stale manifest across a reboot.** Same shape: reads as live rather than wrongly dead.

### DIFF
--- a/.changes/unreleased/Enhancement or New Feature-20260808-120000.yaml
+++ b/.changes/unreleased/Enhancement or New Feature-20260808-120000.yaml
@@ -1,0 +1,3 @@
+kind: Enhancement or New Feature
+body: "The run manifest now records the pid and hostname of the process executing the workflow. A run killed outright -- SIGKILL, OOM, power loss -- never gets to write a terminal status, so `.agent_status.json` kept claiming `running` forever and the VS Code sidebar spun indefinitely. Readers can now ask the OS whether that process still exists and report the actions as interrupted instead. The check is skipped when the manifest predates this field or was written on another host, so nothing is relabelled on a guess."
+time: 2026-08-08T12:00:00Z
--- a/agent_actions/workflow/managers/manifest.py
+++ b/agent_actions/workflow/managers/manifest.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import hashlib
 import json
 import logging
 import os
@@ -27,6 +28,11 @@ class DuplicateActionError(ConfigurationError):
 
 MANIFEST_SCHEMA_VERSION = "1.0"
 MANIFEST_FILENAME = ".manifest.json"
+
+
+def host_fingerprint() -> str:
+    """Stable per-machine id that does not disclose the machine's name."""
+    return hashlib.sha256(socket.gethostname().encode()).hexdigest()[:16]
 
 
 class ManifestManager:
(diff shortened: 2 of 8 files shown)


Muizzkolapo:
**Deferred, not fixed:** two sibling readers show the same stale spinner and could now be reconciled with the data this PR adds.

- `agent_actions/cli/status.py:25-51` — `agac status` reads only `.agent_status.json` and prints `running` raw.
- `agent_actions/tooling/docs/frontend/lib/transformers.ts:37-43` — the docs site maps `manifest.status === 'running'` to a live workflow with no liveness check.

Both are the same root cause as the sidebar bug, and `pid`/`host_id` now make them fixable. Left out of this PR to keep it to one surface; worth a follow-up.

Muizzkolapo:
**Gate record** (harness, HIGH tier):

- **RED → GREEN** re-recorded after each correction. Final: **8171 Python tests**, ruff + ruff-format + mypy clean; **60 extension tests**; `npm run typecheck` clean for `vscode-extension/src`; `npm run build` succeeds.
- **Review:** `YES_WITH_ISSUES` at `96d6c3a1`. One blind lens returned NO on the first pass and was right — the probe whitelisted `EPERM` and called every other outcome dead, which would have relabelled a live run as interrupted on Windows. Fixed and mutation-checked.
- **Smoke: skipped, justified.** `risk.sh` reports `smoke_required=no` — the diff touches no path in `harness/config/smoke-surfaces.txt`. The HIGH tier comes from the coordinator chokepoint (which is reviewed), not from a smoke surface. Covered instead by the full unit suite, the extension suite, typecheck, build, and `vsce ls` confirming no test or source files ship in the `.vsix`.

